PR #5, opened 2023-06-16
Compression Prototype

voltrevo:
Includes both account-level compression and compressing the data for the entry point.

Details to come.


compression/src/VLQ.sol
```diff
@@ -0,0 +1,83 @@
...
+ * 0x00: 0
+ * 0x01: 1
+ * 0x7f: 127
+ * 0x8100: 128
+ * 0x8105: 133 = 1 * 128 + 5
+ * 0x838005: 49,157 = 3 * 128^2 + 0 * 128^1 + 5 * 128^0
+ * 
+ * https://en.wikipedia.org/wiki/Variable-length_quantity
+ */
+library VLQ {
```
jzaki, line 26: Should general, or 3rd party, libraries be in their own sub-folder?

compression/src/DemoAccount.sol
```diff
@@ -0,0 +1,98 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity >=0.7.0 <0.9.0;
+pragma abicoder v2;
+
+import {WaxLib as W} from "./WaxLib.sol";
+import {IDecompressor} from "./decompressors/IDecompressor.sol";
+
+contract DemoAccount {
```
jzaki, line 8: Demo code could be separate from other lib code, although this will occur once the lib contracts are separated.

jzaki (approved):
Good to have this in `main`

PR #15, opened 2023-07-25
Add webauthn project

JohnGuilding:
This PR adds a new foundry project with submodules and 3 solidity contracts.

**Changes**:
- Adds a new foundry project

- Adds submodules
Added submodules to top level `.gitmodules`. Not sure on best practice here when we're going to have multiple foundry projects with their own submodules.

- Adds basic webauthn solidity contracts
    - Adds 2 third-party libraries from Fresh Crypto Lib: `FCL_elliptic.sol` & `FCL_Webauthn.sol`
    - Adds 1 regular contract which uses the libraries: `Webauthn.sol`.

## Fresh Crypto Lib (FCL)
- Fresh Crypto Lib is an open source library written by Ledger which includes an optimised implementation of sec256r1(P256) (`FCL_elliptic.sol`), and an implementation of WebAuthn2/FIDO2 authentication (`FCL_Webauthn.sol`).
- [Repo](https://github.com/rdubois-crypto/FreshCryptoLib/tree/master), [paper](https://eprint.iacr.org/2023/939.pdf), [EthCC talk](https://www.youtube.com/watch?v=Rlq21oA_FA8)

### Why Fresh Crypto Lib (FCL)?
- **Built by a team with cryptography expertise:** When using open source cryptography libraries, we want to be sure that the code has been written by teams with sufficient expertise. Fresh Crypto Lib was built by the team at Ledger who are an established team and they have cryptography expertise within their team.
- **Gas efficiency:** a big problem with secp256r1 curve verification is that it's not supported in the EVM and solidity implementations consume a lot of gas. Fresh Crypto Lib is looking like one of the most (if not the the most) gas efficient open source implementation out there in the wild. They even wrote a paper about it (linked above).
- **MIT licensed** 

Currently unsure if FCL has been audited. Couldn't find reference to an audit in their repo, paper or EthCC talk. It is also difficult to tell whether other open source secp256r1 implementations have been audited. In Ledger's talk, they did [outline several vulnerabilities in other implementations](https://youtu.be/Rlq21oA_FA8?t=659), so that gives me more confidence they have thought about security.

Using FCL is also just an initial option. It seems like the best option at the moment considering security and gas efficiency. We have the option to change implementation if we need to. For example, we may want to integrate the work of [zkwebauthn](https://github.com/zkwebauthn) which is a grantee team in PSE when their work is ready.


webauthn/src/EllipticCurve.sol
voltrevo, line 1: Similarly, it looks like this comes from https://github.com/tdrerup/elliptic-curve-solidity with some light changes to make it work with solidity 0.8. Could you add a comment to that effect?

If this is a change that we made we should consider doing it in a fork so we can import it and possibly contribute it back to the original project.

webauthn/src/WebAuthn.sol
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
```
voltrevo, line 2: Apache-2.0 is an unusual choice for us. Did this code come from somewhere?

voltrevo (changes requested):
In addition to the inline comments, it looks like something went wrong with the way forge-std has been included. There's a copy of the library instead of just the submodule.

I've uploaded a fix for this here:
https://github.com/getwax/wax/tree/add-webauthn-project--remove-forge-std-copy

It removes the copy from the actual history (1MB of junk otherwise left in), so you'd need to do a hard reset to use it:

```sh
git checkout -b backup

# (Add any uncommitted changes, you might need to `cd` up
# to the root of the repo before doing this to get them all)
git add .

git commit -m 'wip'
git checkout -
git fetch origin

# Double check the output is ffb1..a864. The reason is that
# `add-webauthn-project--remove-forge-std-copy` is a replacement
# for that commit, so when you do the hard reset any additional work
# won't be included. If there's additional work, you should be able to
# cherry-pick it over though.
git rev-parse HEAD

git reset --hard origin/add-webauthn-project--remove-forge-std-copy
git push -f
```

(I've added extra commands and instructions to save your work in a backup branch. Always be careful with `git reset --hard`.)

If you need to do cherry-picking to include additional work (as described above), you can do it with something like this:

```sh
git cherry-pick ffb16e64190ab0f98cbaea3ca5ddcf4d547fa864..backup
```

Happy to help if you encounter any git weirdness.

webauthn/src/WebAuthn.sol
JohnGuilding: Got it from somewhere else. Changed file and changed license to MIT

JohnGuilding:
> In addition to the inline comments, it looks like something went wrong with the way forge-std has been included. There's a copy of the library instead of just the submodule.

@voltrevo completely missed this, good catch. Used the fix from your branch and worked without any problems. Thinking back, I initialised the foundry project with the `--no-git` flag. That was likely the cause of the large diff and was a misunderstanding of the flag on my part.

webauthn/src/Base64URL.sol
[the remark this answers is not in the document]
JohnGuilding: I originally got that contract from [here](https://github.com/alembic-tech/P256-verify-signature/tree/main/contracts ) and also saw it used [here](https://github.com/btchip/Webauthn.sol).

Agreed on the point about adding a comment, but have now switched out the contracts used. The new contracts properly cite the author and use openzeppelin's base64 implementation, instead of the more custom base64url implementation.

JohnGuilding:
@voltrevo good points about citing and licenses for the new contracts. My aim was to use those contracts as an initial base to write some tests off and integrate into a Safe plugin. They were third-party contracts I had seen being used in some open source repos using passkeys. I had been planning on potentially replacing them after I had written some tests and integrated them into a plugin. I was familiar using them for local tests so thought they would be a good initial option to get a POC working.

I've decided to swap the contracts out for some contracts from a library called Fresh Crypto Library (FCL). I've updated the description with some context about it. I saw a talk at EthCC where FCL was presented for passkeys/webauthn and seems like a good option to go for, and better than the contracts I had before.

I've also recently been speaking to someone from [Rhinestone](https://twitter.com/rhinestonewtf) who has built some forge tests using the FCL libraries. He's shared them with me so should be easier to build some tests off these libraries now.

TLDR: Reasons for changing the PR is you pointing out the licenses being weird for us; now having a testing base for the new contracts courtesy of rhinestone; and then the FCL contracts being better than the ones before. Hopefully that explains my thought process for why I went for those contracts initially, and then why I swapped them out for FCL.

blakecduncan (approved):
💯 lgtm!

voltrevo (approved):

PR #16, opened 2023-07-31
Updates to inpage #1

### LINKED ISSUE #17: Updates to in-page wallet #1

voltrevo:
Resolves #17.

- Make `eth_requestAccounts` return an array instead of a single address (see `eth_requestAccounts` in [EIP-1102](https://eips.ethereum.org/EIPS/eip-1102))
- Implement `eth_accounts`
  - Returns array of already-connected accounts (doesn't need the user's permission)
  - Enables ethers' `getSigner` method
- Use localstorage so you don't need to reconnect your address when you return/refresh
- Use the `eth_accounts` API in the demo to avoid needing to re-request accounts
- Add `disabled` state for `<Button/>` and apply it when an async callback is in progress

### DIFF
--- a/inpage/demo/App.tsx
+++ b/inpage/demo/App.tsx
@@ -68,9 +68,10 @@ const App = () => {
               });
 
               // TODO: Better type information for EthereumApi
-              assert(typeof response === 'string');
+              assert(Array.isArray(response));
+              assert(typeof response[0] === 'string');
 
-              setAddress(response);
+              setAddress(response[0]);
             }}
           >
             Connect
--- a/inpage/demo/App.tsx
+++ b/inpage/demo/App.tsx
@@ -17,6 +17,18 @@ const App = () => {
   useEffect(() => {
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
     (async () => {
+      if (!address) {
+        const accounts = z.array(z.string()).parse(
+          await demo.ethereum.request({
+            method: 'eth_accounts',
+          }),
+        );
+
+        if (accounts.length > 0) {
+          setAddress(accounts[0]);
+        }
+      }
+
       if (address) {
         setBalance(await demo.provider.getBalance(address));
       }
--- a/inpage/demo/App.tsx
+++ b/inpage/demo/App.tsx
@@ -75,15 +75,9 @@ const App = () => {
           style={{ display: 'inline-block' }}
           type="button"
           onPress={async () => {
-            // TODO: Better type information for EthereumApi
-            const response = z
-              .array(z.string())
-              .min(1)
-              .parse(
-                await demo.ethereum.request({
-                  method: 'eth_requestAccounts',
-                }),
-              );
+            const response = await demo.ethereum.request({
+              method: 'eth_requestAccounts',
+            });
 
             setAddress(response[0]);
           }}
--- a/inpage/demo/App.tsx
+++ b/inpage/demo/App.tsx
@@ -6,6 +6,8 @@ import DemoContext from './DemoContext';
 import Heading from '../src/Heading';
 import assert from '../src/helpers/assert';
 
+const globalRecord = globalThis as Record<string, unknown>;
+
 const App = () => {
   const demo = DemoContext.use();
 
@@ -78,6 +80,15 @@ const App = () => {
           </Button>
         </div>
       )}
+      <Button
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        onClick={async () => {
+          const signer = await demo.provider.getSigner();
+          globalRecord.signer = signer;
+        }}
+      >
+        window.signer
+      </Button>
     </>
   );
 };
--- a/inpage/demo/App.tsx
+++ b/inpage/demo/App.tsx
@@ -63,8 +63,7 @@ const App = () => {
           <Button
             style={{ display: 'inline-block' }}
             type="button"
-            // eslint-disable-next-line @typescript-eslint/no-misused-promises
-            onClick={async () => {
+            onPress={async () => {
               // TODO: Better type information for EthereumApi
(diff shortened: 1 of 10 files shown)


inpage/src/Button.tsx
```diff
@@ -39,24 +42,76 @@ const sheet = jss.createStyleSheet({
...
+          await onPress(e);
+          setError(undefined);
+        } catch (newError) {
+          // eslint-disable-next-line no-console
+          console.error(newError);
+          setError(newError);
+        }
+
+        setLoading(false);
+      })().catch(() => assert(false));
```
JohnGuilding, line 88: What is the purpose of `assert(false)`?

inpage/demo/App.tsx
```diff
@@ -15,6 +17,18 @@ const App = () => {
   useEffect(() => {
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
     (async () => {
+      if (!address) {
+        const accounts = z.array(z.string()).parse(
+          await demo.ethereum.request({
+            method: 'eth_accounts',
```
JohnGuilding, line 23: To clarify the difference between `eth_accounts` and `eth_requestAccounts`, `eth_accounts` only returns an array of **already connected** accounts, whereas `eth_requestAccounts` requests permission from the user to get all accounts (assuming the accounts not connected)?

inpage/src/WaxStorage.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export type WaxStorage = {
+  [K in keyof typeof schema]: Field<z.infer<(typeof schema)[K]['type']>>;
+} & {
+  clear(): Promise<void>;
+};
+
+export default function makeLocalWaxStorage() {
+  const fields = mapValues(schema, ({ type, default_ }, key) => ({
```
JohnGuilding, line 27: I'm going to try explaining this to make sure I understand how `mapValues` is being used here, please correct me if I'm wrong:

We want to store certain values in local storage such as `connectedAccounts` so we can persist state when refreshing the page or leaving and coming back. We can call `makeLocalWaxStorage` to create an object that can operate on localStorage using the methods `get`, `set` and `clear`. And the point of this is to make dealing with local storage easier and add some type safety with zod?

inpage/src/EthereumApi.ts
```diff
@@ -1,43 +1,118 @@
...
+    params: emptyParams,
+    output: z.array(z.string()),
+  },
+};
+
+type Schema = typeof schema;
+
+type RequestParams<M extends string> = M extends keyof Schema
+  ? undefined extends z.infer<Schema[M]['params']>
+    ? { params?: z.infer<Schema[M]['params']> }
```
JohnGuilding, line 25: Not overly familiar with zod, what's the inner ternary doing here? Is it checking if `params` is optional, then making the RequestParams type reflect that?

inpage/demo/App.tsx
```diff
@@ -1,10 +1,12 @@
+import z from 'zod';
```
JohnGuilding, line 1: Never used zod before but heard good things!

JohnGuilding (approved):
Looks good to me!

Added some non-blocking questions for my understanding

inpage/src/WaxStorage.ts
voltrevo: Yep, exactly.

inpage/src/EthereumApi.ts
voltrevo: [zod](https://zod.dev/) is a runtime typing library for TypeScript. It's a better version of io-ts that you may have seen previously in bls-wallet. I believe Zod has since become the defacto standard for doing this kind of thing.

Here's a video about zod: https://www.youtube.com/watch?v=9N50YV5NHaE.

> Is it checking if params is optional, then making the RequestParams type reflect that?

Pretty much. The issue is that if `z.infer<Schema[M]['params']>` is `[] | undefined`, then without the ternary thing we'll end up with `{ method: 'eth_accounts', params: [] | undefined }`. That's a problem because the user will have to write `ethereum.request({ method: 'eth_accounts', params: undefined })` instead of just `ethereum.request({ method: 'eth_accounts' })`.

I hope someday we can use more regular syntax for expressing type logic. I imagine something like this would make it a lot clearer:

```ts
generic RequestParams(M: string) {
  if (!(M extends keyof Schema)) {
    return { params?: unknown[] };
  }

  type Params = z.infer<Schema[M]['params'];

  // Allow omitting the `params` field if it can be `undefined`
  if (undefined extends Params) {
    return { params?: Params };
  }

  return { params: Params };
}
```

inpage/src/Button.tsx
voltrevo: If the async function throws an exception, then the `(async () => { ... })()` expression simply evaluates to a rejected promise. This is something that eslint is picking up. Adding `.catch(() => assert(false))` throws an exception.

It's probably not ideal though, since modern JS runtimes will report uncaught exceptions in promises anyway. I'll take a look at this and maybe just disable the rule.

inpage/demo/App.tsx
voltrevo: Pretty much. A couple of notes about `eth_requestAccounts` though:

- If any accounts are already connected, then `eth_requestAccounts` will just return those accounts without re-prompting (this might be just a convention, the requirement is just that `eth_requestAccounts` returns a non-empty array if it succeeds)
  - (In other words `eth_requestAccounts` only prompts when needed)
- `eth_requestAccounts` is only asking for access to accounts that the user is willing to grant, which is usually a single account, it's not asking for *all* accounts

inpage/demo/App.tsx
voltrevo: Yeah! 🔥

PR #22, opened 2023-08-01
Inpage: Deploy contracts and add Greeter dApp

### LINKED ISSUE #18: Integrate Greeter contract into in-page wallet

voltrevo:
Resolves #18.

# Video Walkthrough

[Video walkthrough](https://www.youtube.com/watch?v=DXLgWCnLFns)

Around 22:20 I said I might record a supplemental video about why I didn't use ethers.formatEther. I remember the reason now is quite simple - ethers.formatEther enforces at least one decimal place, which means it goes beyond the specified number of significant figures in some cases.

# `waxInPage.getContracts()`

`waxInPage.getContracts()` can now be used to get `Greeter`, `EntryPoint`, and `SimpleAccountFactory`:

Screen Shot 2023-08-01 at 4 55 10 pm

If those contracts have not yet deployed you'll get a popup like this and `waxInPage` will deploy the contracts using the provided seed phrase or private key before resolving the `.getContracts()` call.

Screen Shot 2023-08-01 at 5 01 22 pm

`SafeSingletonFactory` is used to deploy everything so if the contracts are already deployed you'll just connect to the same ones without a prompt. Also if the previous deployment is a different version it'll surface as `waxInPage` trying to deploy the correct version.

In a deployed dApp, you can do `waxInPage.setConfig({ deployContractsIfNeeded: false });` to prevent accidentally showing this UI. If the contracts aren't found then `.getContracts()` will throw an error instead.

# Also

Other stuff going on, most notably the greeter dapp. See the diff or video walkthrough for full changes.


blakecduncan (approved):
LGTM!  Watched the demo video, and that was very informative.  Thanks for recording it.  No real comments/concerns from me.  I think it looks good, and I like the way it's progressing.  Also, I'm cool with you using your own logic instead of libraries.  I haven't done a deep dive into the react libraries in a while, so I'm not pushing to swap anything out.

PR #27, opened 2023-08-01
Create base Foundry & Hardhat from template

jzaki:
No other commits on top of the template yet.


aa-plugin-dev/.github/workflows/test.yml
```diff
@@ -0,0 +1,64 @@
+name: test
+
+on:
+  push:
+    branches:
+      - master
```
JohnGuilding, line 6: ~Not sure if we want to update this now, but if we do, should this use `main` instead of `master`?~

**Edit**: Seen your comment about no other commits on top of template just yet, so assuming changes like this would come in a subsequent PR

aa-plugin-dev/package-lock.json
JohnGuilding, line 1: Looks like this template uses npm instead of yarn as a package manager. I know the inpage project uses yarn and we've also used yarn in the past, so not sure if we want to stick with yarn for consistency. wdyt?

I'm also not sure if Lerna works well with a monorepo with yarn and npm projects together. I couldn't see any useful info after a quick skim of their docs. @jacque006 do you know if this is possible?

JohnGuilding (changes requested):
Looks like you ran into the same issue I did when adding a foundry project. It appears that the `forge-std` and `solmate` submodules have been copied instead of being added as submodules. @voltrevo pointed out a similar issue with my PR, you can see his [comment here](https://github.com/getwax/wax/pull/15#pullrequestreview-1553757209).

Adding a submodule should just be a single line change like this:
Screenshot 2023-08-02 at 12 48 32

The reason for me was I incorrectly used the `--no-git` flag when initialising the foundry project. Not sure what happened here though. Andrew's comment might be helpful, but otherwise not sure why the submodules would be copied over like this.

PR #33, opened 2023-08-07
Add a BLS plugin for a Kernel scw

blakecduncan:
Creating a verification plugin for Kernel that uses BLS signatures.  Currently have one test to verify the signature works.  The bls signature is created off-chain (in a separate typescript script) and hardcoded into the test.  I wasn't sure how to create the signature in solidity or if it was even possible.


kernel-plugins/src/BLSValidator.sol
```diff
@@ -0,0 +1,75 @@
...
+
+// BLSValidator is a validator that uses BLS signatures to validate transactions.
+// TODO: Consider account recovery, aggregate signatures, and use EIP 712.
+// Note: This is not audited and should not be used in production.
+contract BLSValidator is IKernelValidator {
+    bytes32 public constant BLS_DOMAIN = keccak256("eip4337.bls.domain");
+    
+    event OwnerChanged(address indexed kernel, uint256[4] indexed oldPublicKey, uint256[4] indexed newPublicKey);
+
+    mapping(address => uint256[4]) public blsValidatorStorage;
```
JohnGuilding, line 18: What is this mapping for exactly? It has quite a generic name but appears to be a mapping of owners to bls public keys.

Is it to do with 4337 storage rules? https://docs.zerodev.app/extend-wallets/build-a-plugin#validator-storage

kernel-plugins/src/BLSValidator.sol
```diff
@@ -0,0 +1,75 @@
...
+            return 0;
+        }
+        // TODO: check if wallet recovered
+        return SIG_VALIDATION_FAILED;
+    }
+
+
+    function validateSignature(bytes32 hash, bytes calldata signature) public view override returns (uint256) {
+        uint256[4] memory publicKey = blsValidatorStorage[msg.sender];
+        uint256[2] memory decodedSignature = abi.decode(signature, (uint256[2]));
```
JohnGuilding, line 60: This might be overcomplicating things for an initial PR, but was just cross-referencing this with code in the v1 verification gateway. Do we want to add checks like the following at this stage?
```solidity
require(signature.length == 64, "Sig bytes length must be 64");
```

Or if we want to move to custom errors:
```solidity
if (signature.length != 64) revert SIG_BYTES_LENGTH_MUST_BE_64();
```

Same with checking `publicKey` is not `[0,0,0,0]`.

Again, might be overcomplicating things for now but thought I'd ask just in case.

kernel-plugins/test/foundry/BLSValidator.test.sol
```diff
@@ -0,0 +1,181 @@
...
+        bytes memory enableData = abi.encodePacked(
+            publicKey,
+            uint48(0),
+            TestExecutor.doNothing.selector,
+            uint48(0),
+            uint48(0),
+            uint32(16)
+        );
+
+        bytes32 digest = getTypedDataHash(
```
JohnGuilding, line 69: Is `getTypedDataHash` just needed for the ECDSA sig?

kernel-plugins/test/foundry/BLSValidator.test.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint256[2] memory signature = [
+            0x238bd40a43095cc868aa13820ced70e0deedc62f588a029cd9aad8a6b72a2283,
+            0x21f8022a28193d60e8f28bb4ce35086c28b1bbe5bb5bf1564bc7b63a2e544f54
+        ];
+
+        // First signature is the ecdsa signature for the default (ecdsa) validator. We
+        // need that signature to validate the BLSValidator for the kernel account.
+        // Second signature is the BLS signature for the BLSValidator. We are calling
+        // the TestExecutor with the BLSValidator as the validator.
+        op.signature = bytes.concat(op.signature, abi.encodePacked(signature));
```
JohnGuilding, line 106: Can you explain why we need to create `op.signature` and then concat it with the hard-coded bls sig? Not sure I'm quite understanding the kernel flow

JohnGuilding (approved):
Looks good to me 🚀

Added a few non-blocking questions/comments

kernel-plugins/remappings.txt
```diff
@@ -0,0 +1,7 @@
+@openzeppelin/=lib/kernel/lib/openzeppelin-contracts/
```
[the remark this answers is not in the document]
blakecduncan: Those were the remappings that were auto-generated from this command.

```
forge remappings > remappings.txt 
```
It does look like some of those are unnecessary.  I'm wondering if that went into the submodules and found remappings there as well.  I'll remove the unnecessary ones cause it looks like tests still pass.

kernel-plugins/src/BLSValidator.sol
blakecduncan: It is a mapping of the address of the kernel wallet to a BLS public key.  So a transaction coming from that kernel wallet can be verified with the public key in the mapping.

Yeah not sure if this is how it is with `Safe` but the Validator plugins with kernel are link singletons in my mind. So several kernel wallets can use the same validator rather than a validator being deployed for each kernel wallet.  Acting similarly to our verification gateway.

But yeah is it to do with the storage rules, I copied the structure of the ECDSA validator that zerodev wrote.

kernel-plugins/src/BLSValidator.sol
blakecduncan: No I think that's a good call.  It took me so long just to get something working that I forgot about this.  I can probably go back now and add some validation/checks

kernel-plugins/test/foundry/BLSValidator.test.sol
```diff
@@ -0,0 +1,181 @@
...
+        // Note: With the Kernel wallet validation and execution are seperate
+        // Executors are plugins that add custom functions to Kernel and each function
+        // is tied to a validator.  In this test we will be enabling the blsValidator and
+        // tying the TextExecutor to it.
+        TestExecutor testExecutor = new TestExecutor();
+        UserOperation memory op =
+            entryPoint.fillUserOp(address(kernel), abi.encodeWithSelector(TestExecutor.doNothing.selector));
+
+        // BLS public/private key pair and signatures created using the '@thehubbleproject/bls' library.
+        // Hard coded in the test because I don't know of a way to do this in solidity.
```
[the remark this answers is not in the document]
blakecduncan: This is a good find, I'll definitely check this out.  It was very annoying having to create the signature outside of the test.

kernel-plugins/test/foundry/BLSValidator.test.sol
blakecduncan: I need to look into this a little more, I think something like this is needed to make the EIP-712 message but I'm not 100% sure.  Right now, my domain is just 

```
bytes32 public constant BLS_DOMAIN = keccak256("eip4337.bls.domain");
```

which is also not good for EIP-712.  I added a comment at the top of the validator plugging to look into EIP-712.

kernel-plugins/test/foundry/BLSValidator.test.sol
blakecduncan: Yeah, I copied another test for this, but I'll try my best to explain it.  When you enable a validation plugin for a kernel wallet, you associate that validation with an execution function.  After doing so you can call that execution function from your kernel wallet with the associated validation method.

In this test I'm doing the following:
- Create a UserOp to call the `TestExecutor.doNothing` function from my existing kernel wallet
- Pass the `blsPublicKey` and the `TestExecutor.doNothing.selector` in the enable data.  This will associate the blsPublicKey with this function.
- enabledata, validator, and executor are signed with a ECDSA signature and set to the `op.signature` value.

I think at this point we have basically created an ecdsa signature that is enabling the new validator.  

- Then we concat the ecdsa sig and a BLS signature of the userOp hash.
- This BLS signature will be used to call `TestExecutor.doNothing` after the BLSValidator is enabled with the ecdsa signature.

blakecduncan: I need to see if I can find some documentation on this structure.  I basically went off of how other tests were set up and just figured it out.  I can also create more tests to test out other situations.  For example, can we enable a validator without calling the execution function?  Also, I want a test of calling to execution function multiple times after enabling the validator.

kernel-plugins/remappings.txt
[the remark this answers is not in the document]
JohnGuilding: got it, yeah I've ran that command a few times and it's also given me some incorrect values

kernel-plugins/test/foundry/BLSValidator.test.sol
```diff
@@ -0,0 +1,181 @@
...
+
+        validator = new ECDSAValidator();
+        ecdsaFactory = new ECDSAKernelFactory(factory, validator, entryPoint);
+
+        kernel = Kernel(payable(address(ecdsaFactory.createAccount(owner, 0))));
+        vm.deal(address(kernel), 1e30);
+        beneficiary = payable(address(makeAddr("beneficiary")));
+    }
+
+    function test_mode_2_bls() external {
```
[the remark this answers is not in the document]
JohnGuilding: right ok that makes sense, thanks for the explanation!

kernel-plugins/src/BLSValidator.sol
JohnGuilding: ok got it, that makes sense regarding the plugin as a singleton, and explains why a mapping would be used

kernel-plugins/test/foundry/BLSValidator.test.sol
JohnGuilding: yeah my understanding of EIP 712 isn't too good either. Only have a very high level understanding of it so need to look deeper into it too

kernel-plugins/test/foundry/BLSValidator.test.sol
JohnGuilding: Ok interesting. I think that flow makes sense as you've explained it.

The ECDSA sig part feels quite weird to me. Few follow up questions to clarify:
1. Is the ECDSA signer equivalent to the owner of the kernel wallet?
2. Could you in theory swap out the ECDSA sig for another sig? Like a webauthn sig, or another bls sig?
3. Does the ECDSA signature need to be done only to enable/disable the validator? Or is it always needed to send a user op using the validator too?
4. Is the execution function a singleton too? Or is that specific to an account?

blakecduncan: Good questions - I'll try to answer them:

1. Yeah.  The default validator for the new Kernel wallet in my tests is an ECDSA validator.  So whoever owns that private key owns that Kernel wallet.
2. Yeah, I'm only using the ECDSA sig because it is the default validator when I created the wallet.  I use an ECDSAKernelFactory to generate the wallet in the test.  Theoretically, I could create a WebAuthNKernelFactory that creates the kernel with a webauthn validator as the default validator.

Some good notes from the documentation on default validators
> Each Kernel account must have one "default validator." A default validator is special in a few ways:
> - It's the validator used in "sudo mode" (see overview). In other words, when other validators are not used, the default validator is used by default (duh).
> - It's the validator used in "enable mode" to check if it approves of the enabling of another validator.
> - It's the validator used for checking signatures for ERC-1271.
> The second and third points are the reasons why a default validator must implement a validateSignature method.

3. It is just needed to enable/disable the validator (I should also test out disabling a validator).  Once the validator is set for an executor, you just need the applicable signature.
4. I think executors can be anything you want. You can use an existing executor or deploy your own.  It is just like adding a custom function to your wallet.  The only requirement is that executors don't touch any storage used by the Kernel wallet.  Here's an example they created for erc721 functions https://github.com/zerodevapp/kernel/blob/main/src/executor/TokenActions.sol

JohnGuilding: ok all makes sense, thanks for the detailed answers!

PR #41, opened 2023-08-14
Update readme (and restructure repo)

jzaki:
Changes:
- Update readme. Communicates the why/what of the repo, and how others can benefit
- Addition to top-level .gitignore
- Restructure repo as documented

To ensure the restructure didn't break sub-folders, please do a quick confidence check of using these components:
- [x] @JohnGuilding tested `webauthn`
- [x] @blakecduncan tested `kernel-plugins`
- [x] @voltrevo tested (and or will remedy) `inpage`, `compression`

(we shouldn't have had relative paths between, but just to be sure)


.gitignore
```diff
@@ -141,8 +141,11 @@ out/
 /broadcast/*/1337/
 /broadcast/**/dry-run/
 
-# Docs
-docs/
+# drawio backups
+docs/images/.$*.bkp
```
JohnGuilding, line 145: 1. Does the draw.io backup need to be added to a gitignore? Or is this just something on your machine?
2. There is also an argument for it being kept in source control, otherwise changes might always rely on you having your local copy. If the backup was in source control, any team member could make changes as/when needed. wdyt?

docs/images/wax-composition.svg
JohnGuilding, line 1: 1. Thoughts on having a key on the diagram for each colour instead of/as well as using text? Might make it slightly more visual, and would make it easy to reuse the diagram if we ever needed it somewhere else like a presentation
2. What does the blue mean in the diagram? Can't see it defined anywhere
3. Nitpick - the arrows on the diagram look a bit squished in places e.g. in-between boxes titled “Recover via email” and “zkEmail (email content)”. Also some of the text is overflowing in places
4. What is your thinking behind Obscuro? It doesn’t seem that well established. Are we doing some sort of partnership with them? Otherwise seems like quite a niche choice for a diagram
5. Why is the BLS flow only shown for kernel? It sort of reads as it would only sit there, as opposed to being used in other integrations too

README.md
```diff
@@ -1,2 +1,38 @@
...
+- Within the EthDK there will be additions that enable projects to readily adopt novel features. 
+- On-chain, between smart contracts and primitives, the verification data+logic will be designed to be composable. Smart accounts can then dynamically benefit from more than one primitive, and this will be developed in sync with broader smart account modularisation.
+
+# Using this monorepo
+## /demos
+You're a web dev and want to bring the latest web3 capabilities to your product/users.
+Here you will not only find 'hello world' wrappers of base primitives, but also composite examples highlighting novel product features.
+
+### EthDK (imported?)
+The EthDK repo is useful for SDK devs who also want to support interaction with the bleeding edge of smart account verification capabilities.
```
JohnGuilding, line 29: “The EthDK repo is useful for SDK devs” - makes it sound like the EthDK is mainly for SDK developers, instead of devs who want to use an SDK. Thoughts on something like “The EthDK repo is useful for devs who…”, or “The EthDK repo is useful for wallet and dApp devs who…”

README.md
JohnGuilding, line 1: README looks good!

What do you think about adding a short one-liner at the top of the readme to succinctly explain what WAX is. Think the first section does a good job of explaining, but the reader has to read a sentence and 3 bullet points to get the high level picture. Maybe something like:  “WAX is a set of production-ready components that helps developers easily utilise cryptographic primitives in smart accounts”

JohnGuilding (commented):
Like the readme and the diagram. 

Left a few thoughts on the diagram and a few other comments/thoughts here and there. Let me know what you think!

.gitignore
jzaki: The program that I'm using (draw.io) sometimes makes one, but I'll just get rid of the file and this ignore.

docs/images/wax-composition.svg
jzaki: 1. Added key
2. Blue in key
3. soz
4. Just another privacy preserving component to consider, like Aztec's Noir (also see point 5 below). I'll remove both until we actually do anything on them
5. The arrows are showing what the diagram is atm "current explorations and proposed development", the alternative may be to remove all arrows since the end goal is any-to-any. For now it's just current/upcoming dev.

README.md
jzaki: At this stage "production-ready" is a goal, and I was deliberately leaning away from saying 'cryptographic primitives' on the first line. Trying to start with the benefits for non-web3 devs. I feel web3 devs will be okay to read on if they really want to use it.
We can refine over time.

blakecduncan (approved):

Issue #46, opened 2023-08-15
Add nonce validation in safe plugin


JohnGuilding:
Add nonce validation: https://github.com/getwax/wax/pull/29#discussion_r1291586890

Also check we're overriding the right functions: https://github.com/getwax/wax/pull/29#discussion_r1291585547

PR #48, opened 2023-08-15
Update crypto libs to work with memory variables

JohnGuilding:


webauthn/src/lib/FCL_Webauthn.sol
```diff
@@ -47,55 +48,49 @@ library FCL_WebAuthn {
...
-
-            bytes32 moreData; //=keccak256(abi.encodePacked(challengeExtracted));
-            assembly {
-                moreData := keccak256(add(challengeExtracted, 32), mload(challengeExtracted))
-            }
-
-            if (keccak256(abi.encodePacked(bytes(challengeEncoded))) != moreData) {
-                revert InvalidClientData();
-            }
+            // TODO: wax#44 Extract webauthn challenge from clientData and compare to clientChallenge
```
JohnGuilding, line 51: This logic didn't work out of the box, have added an issue to revisit this at a later stage

webauthn/src/lib/FCL_elliptic.sol
```diff
@@ -834,6 +834,40 @@ library FCL_Elliptic_ZZ {
         return x1 == 0;
     }
 
+
+    /**
+     * @dev ECDSA verification, given , signature, and public key.
+     * @notice Modified from "ecdsa_verify" to use memory instead of calldata
+     */
+    function ecdsa_verify_memory(
```
JohnGuilding, line 842: Take extra care reviewing this change as I had to replace some inline assembly

account-integrations/webauthn/src/lib/FCL_elliptic.sol
```diff
@@ -834,6 +834,40 @@ library FCL_Elliptic_ZZ {
         return x1 == 0;
     }
 
+
+    /**
+     * @dev ECDSA verification, given , signature, and public key.
+     * @notice Modified from "ecdsa_verify" to use memory instead of calldata
+     */
+    function ecdsa_verify_memory(
```
JohnGuilding, line 842: Take extra care reviewing this change as I had to replace some inline assembly

account-integrations/webauthn/src/lib/FCL_Webauthn.sol
```diff
@@ -47,55 +48,49 @@ library FCL_WebAuthn {
...
-
-            bytes32 moreData; //=keccak256(abi.encodePacked(challengeExtracted));
-            assembly {
-                moreData := keccak256(add(challengeExtracted, 32), mload(challengeExtracted))
-            }
-
-            if (keccak256(abi.encodePacked(bytes(challengeEncoded))) != moreData) {
-                revert InvalidClientData();
-            }
+            // TODO: wax#44 Extract webauthn challenge from clientData and compare to clientChallenge
```
JohnGuilding, line 51: This logic didn't work out of the box, have added an issue to revisit this at a later stage

account-integrations/webauthn/src/lib/FCL_elliptic.sol
blakecduncan: I'm also not too familiar with EVM opcodes but your logic/explanation makes sense to me.  I also like your test in the EVM playground.

account-integrations/webauthn/src/SafeECDSAPlugin.sol
```diff
@@ -0,0 +1,91 @@
...
+
+    function validateUserOp(
+        UserOperation calldata userOp,
+        bytes32 userOpHash,
+        uint256 missingAccountFunds
+    ) external override returns (uint256 validationData) {
+        address payable safeAddress = payable(userOp.sender);
+        ISafe senderSafe = ISafe(safeAddress);
+
+        if (missingAccountFunds != 0) {
```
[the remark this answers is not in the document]
JohnGuilding: I think this is replicating [this function](https://github.com/eth-infinitism/account-abstraction/blob/e5658d7a32f8d4e18f38cbf165353691cd7380c7/contracts/core/BaseAccount.sol#L117-L126) in BaseAccount.sol.

I'm also not 100% sure on how things "should" work here given a lot of this code was written by Safe. I made [this issue](https://github.com/getwax/wax/issues/46) to revisit what we should and shouldn't be overriding

account-integrations/webauthn/src/SafeECDSAPlugin.sol
```diff
@@ -0,0 +1,91 @@
...
+        address payable safeAddress = payable(msg.sender);
+        ISafe safe = ISafe(safeAddress);
+        require(
+            safe.execTransactionFromModule(to, value, data, 0),
+            "tx failed"
+        );
+    }
+
+    function enableMyself() public {
+        ISafe(address(this)).enableModule(myAddress);
```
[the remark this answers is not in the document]
JohnGuilding: 1. What is this enableMyself() function?
It's used as a wrapper around the the Safe's `enableModule` function, which is then called from the `setup` function [here](https://github.com/getwax/wax/blob/38368a72a6c8a77db41825c3791ed79018e32edf/account-integrations/webauthn/test/integration/SafeECDSAPlugin.test.ts#L114). The setup function will then call the `setupModules` function [here](https://github.com/safe-global/safe-contracts/blob/af488f0a649a3be2ea01e0f7a6af1587d0250c2d/contracts/base/ModuleManager.sol#L27-L41). I took the function from the safe integration test and if I'm honest, I'm not sure why they chose to create this wrapper function instead of just call `enableModule`. I noticed other modules  using older safe code (when it was called gnosis safe) inherited from [a contract](https://github.com/safe-global/safe-modules/blob/master/recurring_transfers/contracts/RecurringTransfersModule.sol) called `Module`. I would need to look into that further to figure out what is happening under the hood. This is probably something to come back to and assess whether it's needed for our use case in the future, will add an issue for it. 

3. isn't address(this) the same as myAddress
My assumption would be, since this is getting called from the context of the setup function in Safe.sol, `address(this)` is equal to the address of Safe.sol, and not the plugin. Again, not 100% sure on how everything in Safe fits together yet, so would want to verify this at some point

account-integrations/webauthn/src/lib/FCL_elliptic.sol
jzaki: Why is the type uint256 instead of casting a bytes32?
The `sub` assembly function probably doesn't care for the type (just the number of bits), but still curious why have `rs0` a different type to the result of `calldataload`?

Also, what's the reasoning behind the change from `calldata` to `memory`?

account-integrations/webauthn/src/lib/FCL_Webauthn.sol
```diff
@@ -28,13 +28,14 @@ library FCL_WebAuthn {
     error InvalidClientData();
     error InvalidSignature();
 
+    /** @notice Modified from original Fresh Crypto Lib code to use memory instead of calldata */
```
jzaki, line 31: Curious why this change is needed? PS can appreciate that arrays/bytes can be a pain regarding access.

jzaki (approved):

account-integrations/webauthn/src/lib/FCL_elliptic.sol
JohnGuilding: > Why is the type uint256 instead of casting a bytes32?

I was just sticking to the original type. Would you recommend casting to `bytes32`?

> what's the reasoning behind the change from calldata to memory?

The full explanation for why I changed from `calldata` to `memory` is in the PR description. Does it make sense or should I expand upon it?

**TLDR** is that I couldn't find a way to decode the signature without having the variables stored as `memory` because `abi.decode` appears to return `memory` vars. Then without changing FCL, I get the following error when passing these memory vars into the library functions. So I had to change subsequent functions where the vars are used too: 
```
Invalid type for argument in function call. Invalid implicit conversion from bytes memory to bytes calldata requested.
```

I added a snippet in the description which you can try out in remix to test my assumptions. Definitely open to alternatives if anyone knows of any, as I didn't want to make the change

account-integrations/webauthn/src/lib/FCL_Webauthn.sol
JohnGuilding: If referring to using `memory` instead of `calldata`, answered in other comment and PR description. Let me know if what I'm saying doesn't makes sense

PR #52, opened 2023-08-17
28 add initial forge test for webauthn verification

JohnGuilding:
This PR:
- Adds intial forge test for `Webauthn.sol`
- Adds intial forge test for `SafeWebAuthnPlugin.sol`
- Adds `TestHelper.sol` contract to provide helper functions and abstract common logic
- Adds a `SafeWebAuthnPluginHarness.sol` to expose internal functions for testing. This is the recommended way to test internal functions from [the foundry docs](https://book.getfoundry.sh/tutorials/best-practices#internal-functions).

~This PR is dependent on #48~ Merged

PR #54, opened 2023-08-25
Safe BLS module

blakecduncan:
Blocked by: https://github.com/getwax/wax/pull/52

Adding a Safe module that uses BLS sigs to verify a transaction. I added a hardhat test that submits a userOp to the entrypoint directly.  This way we don't need to run a bundler or start a node to test a module.

I'm wondering if we should reorganize our test directory.  Right now we have three forms of tests.
- **hardhat test** - all you need is `yarn hardhat test` and the tests should work. Hardhat tests are useful for verification types that are hard/not possible in solidity.
- **Integration test** - Need to run a bundler and a node separately for the tests to pass.
- **Forge tests** - just need `forge test`, and tests are written in solidity.

Should rename/organise these differently?  For example, the two ideas I have are:

- `test/hardhat`
- `test/forge`
- `test/integration`

or

- `test/hardhat`
- `test/hardhat/integration` (since the integration tests use hardhat)
- `test/forge`

### DIFF
--- a/account-integrations/safe/src/SafeBlsPlugin.sol
+++ b/account-integrations/safe/src/SafeBlsPlugin.sol
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.7.0 <0.9.0;
+pragma abicoder v2;
+
+import {BaseAccount} from "account-abstraction/contracts/core/BaseAccount.sol";
+import {IEntryPoint, UserOperation} from "account-abstraction/contracts/interfaces/IEntryPoint.sol";
+import {BLS} from "account-abstraction/contracts/samples/bls/lib/hubble-contracts/contracts/libs/BLS.sol";
+
+interface ISafe {
+    function enableModule(address module) external;
+
+    function execTransactionFromModule(
+        address to,
+        uint256 value,
+        bytes memory data,
+        uint8 operation
+    ) external returns (bool success);
+}
+
+contract SafeBlsPlugin is BaseAccount {
+
+    bytes32 public constant BLS_DOMAIN = keccak256("eip4337.bls.domain");
+    address public immutable myAddress;
+    uint256[4] private _blsPublicKey;
+    address private immutable _entryPoint;
+
+    address internal constant _SENTINEL_MODULES = address(0x1);
+
+    constructor(address entryPointAddress, uint256[4] memory blsPublicKey) {
+        myAddress = address(this);
+        _blsPublicKey = blsPublicKey;
+        _entryPoint = entryPointAddress;
+    }
+
+    function validateUserOp(
+        UserOperation calldata userOp,
+        bytes32 userOpHash,
+        uint256 missingAccountFunds
+    ) external override returns (uint256 validationData) {
+        address payable safeAddress = payable(userOp.sender);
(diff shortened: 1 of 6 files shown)


account-integrations/safe/src/SafeBlsPlugin.sol
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.7.0 <0.9.0;
+pragma abicoder v2;
```
JohnGuilding, line 3: Seen that this is included in the `SafeECDSAPlugin` too. I believe abicoder was defaulted to v2 as of [solidity 0.8.0](https://docs.soliditylang.org/en/v0.8.21/080-breaking-changes.html#silent-changes-of-the-semantics). I think it makes sense here and in the ECDSA plugin as we're using solidity`>=0.7.0`. Curious if we want to keep supporting versions of solidity less than 0.8.0? Open question don't really have a strong opinion

account-integrations/safe/src/SafeBlsPlugin.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function owner() public pure returns (uint256[4] memory _blsPublicKey) {
+        return _blsPublicKey;
+    }
+
+    function _validateSignature(
+        UserOperation calldata userOp,
+        bytes32 userOpHash
+    ) internal view override returns (uint256 validationData) {
+        require(userOp.signature.length == 64, "VG: Sig bytes length must be 64");
```
JohnGuilding, line 84: What do you think about custom errors as opposed to require statements with strings? https://soliditylang.org/blog/2021/04/21/custom-errors/

account-integrations/safe/src/utils/SafeTypechainTypesHelper.sol
JohnGuilding, line 1: Now that we are importing 4337 contracts, should we rename this to something more generic? Maybe something like `TypesHelper`?

account-integrations/safe/test/hardhat/SafeBlsPlugin.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    const userWallet = ethers.Wallet.fromPhrase(MNEMONIC).connect(
+      provider
+    );
+
+    return {
+        provider,
+        userWallet,
+        entryPoint,
+        safe,
+        safeProxyFactory,
```
JohnGuilding, line 43: If we're creating a `safeProxyFactory` variable and others before the describe block and then assigning to them, is there a reason to return them in this function as well?

JohnGuilding (approved):
Looks good - Approving!

Left a few Qs

account-integrations/safe/src/SafeBlsPlugin.sol
blakecduncan: Sorry just getting around to these comments.  Yeah, I don't have a preference either.  I haven't really done a deep dive to know if it's still necessary to support the older versions

account-integrations/safe/src/SafeBlsPlugin.sol
blakecduncan: Good call, I'll make the update

PR #59, opened 2023-08-29
58 deterministic deploy of contracts in safe project

JohnGuilding:
Dependent on #57 (`28-follow-up`)

Deterministically deploys Safe contracts so we don't have to use new addresses each time we run integration tests

- Uses `SafeSingletonFactory` from `demos/inpage`
- Uses same helper functions used in `demos/inpage` for `SafeSingletonFactory`
- Updates `deployAll.ts` script to use `SafeSingletonFactory`
- Updates readme & env.example
- Sets `settings` value for hardhat preprocessor in `hardhat.config.ts` - [see comment](https://github.com/getwax/wax/pull/59/files#r1308765134) on why this is needed


account-integrations/safe/hardhat.config.ts
```diff
@@ -48,6 +48,7 @@ const config: HardhatUserConfig = {
         }
         return line;
       },
+      settings: true,
```
JohnGuilding, line 51: **TLDR:**
This is needed to stop hardhat generating new bytecode every time we run the deploy script. New bytecode = different address each time because of the way `create2` works

**Long answer:**
In order for hardhat to play nicely with foundry, we use [hardhat-preprocesser](https://github.com/wighawag/hardhat-preprocessor) as [suggested by the foundry docs](https://book.getfoundry.sh/config/hardhat#integrating-with-hardhat). The preprocessor plugin allows us to modify contract code before compilation. Hardhat expects dependencies to be installed in `node_modules`, so the pre-processor points foundry in the direction of `lib/` instead of `node_modules`.

The issue I ran into with hardhat-preprocessor is it creates a field called `__CACHE_BREAKER__`. It does so if a value called `settings` does not exist. You can see how it works under the hood [here](https://github.com/wighawag/hardhat-preprocessor/blob/dac6c715ce1fa5af83441302d655d8922cfc4880/src/index.ts#L179). if `!linePreProcessor.settings` is true, then a new cache breaker is generated using the current time.

I was getting some weird behaviour where running `hardhat compile` would always compile contracts again even when I had changed nothing in the contracts. Normally you would just get a `Nothing to compile` message. 

So what I think was happening was that changing the cache breaker each time resulted in hardhat creating slightly different bytecode each time we run the deploy script. I tested this by comparing bytecode between hardhat compiles and it differed very slightly towards the end. I haven't figured out exactly how hardhat generates bytecode and how that is related to `__CACHE_BREAKER__`, but have inferred through this change that a different cache breaker leads to new bytecode.

If the byetcode changes, then we get different values inputted into `create2`, which is what the `SafeSingletonfactory`, so we get a different address each time. I hope this roughly makes sense, let me know if anything mentioned is not clear. May update here if I figure out the link between `__CACHE_BREAKER__` and byetcode generation.

account-integrations/safe/test/hardhat/utils/SafeSingletonFactory.ts
JohnGuilding, line 1: I'm open to the idea of sharing this file among projects as opposed to just copying it over from `inpage`. I did try and import it briefly from `inpage`, but got this error when running the script and decided to try copying it for simplicity

Screenshot 2023-08-29 at 14 33 55

blakecduncan (approved):
Nice!  This looks good

PR #65, opened 2023-08-30
46 add nonce validation in safe plugin

JohnGuilding:
Dependent on #63 (`62 add basic GitHub workflows`)

This PR makes the following changes to `SafeBlsPlugin`, `SafeECDSAPlugin` and `SafeWebAuthnPlugin`:

### Validates userOp.nonce is sequential
I've opted for a sequential nonce as I don't see the need for anything more complicated right now. Please shout if you disagree.

There is no need to validate nonce uniqueness as the EntryPoint already does that for us. See section titled "Semi-abstracted Nonce Support" in the [ERC](https://eips.ethereum.org/EIPS/eip-4337). As specified by the ERC, to enforce a classic sequential nonce the validate function should perform something like:
```solidity
require(userOp.nonce<type(uint64).max)
```

### Only overrides internal functions
Instead of overriding the whole of `validateUserOp`, the plugin now only overrides the relevant internal methods. Don't know if this is super necessary but a comment in `BaseAccount.sol` seems [to suggest it](https://github.com/eth-infinitism/account-abstraction/blob/dae9733bf78bcb7576f572d67497c2d241ae4da1/contracts/core/BaseAccount.sol#L42-L43). The slightly annoying thing about this approach is that the `_requireFromEntryPoint();` function will always fail as the function appears to be called from a Safe contract (I think?), rather than the EntryPoint. That is why I had to override `_requireFromEntryPoint()` and do nothing. We could override it to actually do something, but anything that wasn't requiring from the EntryPoint would be misleading.

When running the integration test and not overriding the method, I would get the error `account: not from EntryPoint`. I'm not sure I exactly understand the Safe flow here, but encountered the error and this was my workaround. Open to the idea of overriding `validateUserOp` again if that's less confusing, let me know your thoughts!

### Other comments
- There's a few untested functions in the Safe plugins, created an issue to look into that - https://github.com/getwax/wax/issues/66
- If any of the nonce stuff doesn't make sense like the use of and mention of keys, let me know and will write better comments

### DIFF
--- a/.github/workflows/compression.yml
+++ b/.github/workflows/compression.yml
@@ -1,0 +1,41 @@
+name: account-integrations/compression
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - account-integrations/compression/**
+
+defaults:
+  run:
+    working-directory: ./account-integrations/compression
+
+jobs:
+  check:
+    strategy:
+      fail-fast: true
+
+    name: Foundry project
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+
+      - name: Run Forge build
+        run: |
+          forge --version
+          forge build --sizes
+        id: build
+
+      - name: Run Forge tests
+        run: |
+          forge test -vvv
+        id: test
--- a/.github/workflows/kernel.yml
+++ b/.github/workflows/kernel.yml
@@ -1,0 +1,41 @@
+name: account-integrations/kernel
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths:
+      - account-integrations/kernel/**
+
+defaults:
+  run:
+    working-directory: ./account-integrations/kernel
+
+jobs:
+  check:
+    strategy:
+      fail-fast: true
+
+    name: Foundry project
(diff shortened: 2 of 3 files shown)


account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -75,6 +57,9 @@ contract SafeWebAuthnPlugin is BaseAccount {
         return _publicKey;
     }
 
+    /** @notice validateSignature is called from Safe so the original function will always fail */
```
JohnGuilding, line 60: Adding a comment for myself, verify exactly where validateSignature is getting called from

jzaki (approved):

JohnGuilding:
Closed due to resetting this branch to main. This was because I messed up the git history slightly and changes were showing in this branch that had already been merged into main.

To get this branch to a clean slate I ran `git reset --hard main`. I force pushed to this remote branch to check that worked, because there was no difference between this PR and main, github automatically closed the PR. This was unintentional, as I was planning on cherry picking the correct commits from a local backup branch and then pushing remotely so that this PR would be updated. I think what I should have done is not forced pushed after resetting to main, and only pushing remotely once I'd cherry picked the right commits.

I still cherry picked and pushed the right commits to the remote branch, but it didn't reopen the PR and I raised a new PR instead with the correct commits. For reference, I ran `git cherry-pick 4db2df7^..15ee10f `, which included only the commits that should be showing in this PR

Replacement PR with clean commits https://github.com/getwax/wax/pull/69

Issue #43, opened 2023-08-15
Hello World ZKP Verification


jacque006:
Simplest ZK circuit that can be used for 4337 validation

(strech) Implement with safe

PR #77, opened 2023-09-12
Add ZKP Password Example

jacque006:
### Description

- Add zkp project to integrations with password circuit, client, tests, and zk build toolchain.
- Add zkp password Safe plugin to show integration. Use mocked verifier for now.
- Update start script to include bundler docker container.

### Issues

Resolves https://github.com/getwax/wax/issues/43
Contributes towards https://github.com/getwax/wax/issues/79

### Followup

- https://github.com/getwax/wax/issues/143
- https://github.com/getwax/wax/issues/146
- https://github.com/getwax/wax/issues/147

### TODO
- [x] Resolve merge conflicts
- [x] Track down intermittent test failure
- [x] Link generated circuit assets in github workflow tests
- [x] ~See if circom/circuit artifacts can be done outside of core workflows~ Moved to https://github.com/getwax/wax/issues/143
- [x] Update zkp `README`, including non-production readiness
- [x] Update integration test
- [x] Fill in remaining docstrings


account-integrations/safe/src/SafeZKPPasswordPlugin.sol
```diff
@@ -0,0 +1,109 @@
...
+    ) internal view override returns (uint256 validationData) {
+        // TODO (merge-ok) There is likely a more efficient way to encode this
+        // to save on space, which would be especially desirable on rollups.
+        uint256[2] memory a;
+        uint256[2][2] memory b;
+        uint256[2] memory c;
+        (a, b, c) = abi.decode(
+            userOp.signature,
+            (uint256[2], uint256[2][2], uint256[2])
+        );
```
jacque006, line 92: @voltrevo Curious if your compression work could be applied here for better results.

account-integrations/safe/test/hardhat/integration/SafeZKPPasswordPlugin.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      );
+
+    // Native tokens for the pre-fund 💸
+    await receiptOf(
+      userWallet.sendTransaction({
+        to: accountAddress,
+        value: ethers.parseEther("100"),
+        // Re-sync nonce
+        // TODO (merge-ok) See if NonceManager could fix this.
+        nonce: await userWallet.getNonce(),
```
jacque006, line 134: @JohnGuilding This is the only spot in my tests that had nonce issues. Looks like `userWallet` is desyncing somewhere above.

JohnGuilding: `userWallet` uses the default mnemonic by the look of things. When using the default mnemonic for the admin signer in my changes, I was getting nonce too low errors too, and resolved those with a nonce manager. Imagine wrapping it in a `NonceManager` will resolve this for you too.

It may be clashing with your local bundler if that uses the following mnemonic: `test test test test test test test test test test test junk` (which the default config for eth-infinitism's bundler does). I've just pushed a commit to my PR that avoids using the first derived account from that mnemonic to avoid clashes with  a bundler

account-integrations/safe/src/SafeZKPPasswordPlugin.sol
voltrevo: It depends... looks like not. What does a typical signature look like? If it's basically 8 random u256 values then they aren't really compressible.

Given this is a ZK thing though, perhaps there is a way to aggregate signatures like we do in BLS? (ie hopefully a proof for N ops can take up the same(ish) space as a proof for 1 op?)

jacque006: Yeah I think that intuition about the values is correct. Example contract proof:
```json
{
  a: [
    '17744207585240846412693147604714530875019749677313583372657036954963043981064',
    '1168049317011645485556628350368210451616762497499208564464528439268597146849'
  ],
  b: [
    [
      '16662351504458598424799058929317773366859175778790008142686580172931446126160',
      '7635494043804658990254311186067069547313621379692771118538406803317269268407'
    ],
    [
      '2404765150978297026600341457208632360568319710876326085883565253499494475928',
      '8643448892689514290763234906271149460005611481013939211301224343257746681052'
    ]
  ],
  c: [
    '17571014154693344358582154424375677751358230006768854130079694344334794313411',
    '12483301928936689173606863935123864495226004585408879709965283955283287073047'
  ]
}
```
Assuming there are many, many proofs we may be able to use something like deflate, but would need to run tests to see if space/computation tradeoff is worth it.

Agreed the aggregation of proofs on chain would be a major data/computation saver. I haven't seen anything along those lines yet but the 4337 aggregator entities should be able to support assuming it can be done in EVM.

JohnGuilding (commented):
Looks gooood. README instructions worked well and tests all passed except webauthn hardhat test

JohnGuilding (approved):

account-integrations/safe/src/SafeZKPPasswordPlugin.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.7.0 <0.9.0;
+pragma abicoder v2;
+
+import {IPasswordVerifier} from "./interface/IPasswordVerifier.sol";
+import {ISafe} from "./interface/ISafe.sol";
+import {MockPasswordVerifier} from "./utils/MockPasswordVerifier.sol";
```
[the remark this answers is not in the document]
jacque006: Leftover from when I was hardcoding it in vs passing through constructor. Will remove.

jacque006:
Webauthn test failure is due to bundler -> geth debug call tracing performance. Will link relevant issue here when created.

jacque006:
https://github.com/eth-infinitism/bundler/issues/161

account-integrations/safe/test/hardhat/integration/SafeZKPPasswordPlugin.test.ts
voltrevo: I ran into this too and put this list together so I didn't need to deal with real-looking seedphrases during development: [Test bip39 seedphrases](https://gist.github.com/voltrevo/df7c306d0ccad3bbff3cec95a3ec752a).

Issue #68, opened 2023-09-01
Calldata decoding and stack limit


jzaki:
To avoid modifying the webauthn library, we must setup corresponding param variables to refer to the calldata in the UserOp signature.
There is an as yet unaccounted for slot in decoding.
This work also hits upon a stack limit which needs to be addressed to resume testing the decoded calldata.

PR #83, opened 2023-09-14
Extract and validate client challenge

JohnGuilding:
This PR properly extracts the client challenge and validates it. The reason this wasn't working originally is because the code actually needed to use `base64url` encoding instead of `base64` encoding to encode the `clientChallenge` in `FCL_WebAuthn`.

## Base64URL encoding
The `clientChallenge` with base64url encoding is equal to `NTo-1aBEGRnxxjmkaTHehyrDNX3izlqi1owmOd9UGJ0`. Whereas the base64 encoding of that is `NTo+1aBEGRnxxjmkaTHehyrDNX3izlqi1owmOd9UGJ0=`. This resulted in [the hash](https://github.com/getwax/wax/blob/a86e783dbf0a8d2b8654e8423021e032c0407d81/account-integrations/safe/src/lib/FCL_Webauthn.sol#L79) being incorrect which the extracted challenge was compared to. Therefore the assertion failed and an error was thrown.

### Why did the original ledger implementation not use base64URL encoding?
It turns out the implementation located in `src/` is different to the implementation that is tested against in `tests/`. If you look at the Fresh Crypto Lib (FCL) [implementation](https://github.com/rdubois-crypto/FreshCryptoLib/blob/c32999e145a74e9428ac78a1faed2887ea078e6d/solidity/src/FCL_Webauthn.sol#L23), you can see that it does indeed use `base64` encoding. Since I copied that file originally, I expected that to work. It turns out that the FCL  tests do actually use `base64url` encoding instead! See this link [here](https://github.com/rdubois-crypto/FreshCryptoLib/blob/c32999e145a74e9428ac78a1faed2887ea078e6d/solidity/tests/WebAuthn_hardhat/contracts/FCL_Webauthn.sol#L27).

### Does that mean we always have to use `base64url` encoding now?
TLDR: Not sure. Need to refresh my knowledge of how webauthn handles encoding `clientData`, and whether that is something we can manage. Ideally the aim is to not edit the FCL library contract at all. Added an issue to investigate this further -  https://github.com/getwax/wax/issues/84

~### copyBytes function~
~Since we haven't implemented #68 yet, we couldn't use the original `clientChallenge` validation logic still, as that relied on the arguments being stored as `calldata`. As a temporary solution until we fix that, I've chosen to use a function called `copyBytes`. This is taken from the original webauthn solidity implementation which `FCL_WebAuthn` is [based on](https://github.com/getwax/wax/blob/a86e783dbf0a8d2b8654e8423021e032c0407d81/account-integrations/safe/src/lib/FCL_Webauthn.sol#L15). You can see the original code for that [here](https://github.com/btchip/Webauthn.sol/blob/30ae1795dab8aa3aaf3a3c3c407854121e74c1fc/contracts/Webauthn.sol#L96-L119).~
^ #68 has been merged so we don't need to use the `copyBytes` function anymore

The PR also updates the webauthn sig values used throughout the tests to use a valid clientChallenge (I was using incorrect values such as the userOp hash before)


account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -87,20 +88,22 @@ contract SafeWebAuthnPlugin is BaseAccount {
...
                 s.authenticatorDataFlagMask,
                 , // some number
+                s.clientChallenge,
                 s.clientChallengeDataOffset
             ) = abi.decode(
                 userOp.signature[i:i+dataLen],
                 (
                     bytes1,
                     uint256, //not sure what is encoded here
+                    bytes32,
```
JohnGuilding, line 106: ^I'm not sure why this is the order for the decoding, I originally assumed it would have been bytes before uints like so:
```solidity
(
    bytes1,
    bytes32,
    uint256, 
    bytes32
)
```

The screenshot below shows the difference between the two signatures. The older signature without the `clientChallenge` is on the left, and the newer signature with the `clientChallenge` is on the right:

Screenshot 2023-09-15 at 16 12 30

You can see where the clientChallenge was added, look for `353a3ed5a0441919f1c639a46931de872ac3357de2ce5aa2d68c2639df54189d`. There were some other changes which I'm not too sure about. It looks like the mystery `uint256` changed. The value in the first 32 bytes was also changed, again not sure what that change means - @jzaki any ideas?

account-integrations/safe/src/lib/FCL_Webauthn.sol
```diff
@@ -135,4 +132,4 @@ library FCL_WebAuthn {
 
         return result;
     }
-}
+}
```
jzaki, line 135: I assume this file is now back to it's original version?

jzaki (approved):

account-integrations/safe/src/lib/FCL_Webauthn.sol
JohnGuilding: Not quite, I raised https://github.com/getwax/wax/pull/93 to revert more of the libs back to their original state. Remaining work to restore to it's original state is covered by #84 & #94

PR #85, opened 2023-09-14
Use original library function (calldata params)

### LINKED ISSUE #68: Calldata decoding and stack limit

jzaki:
Slicing calldata for a webauthn library call whilst keeping below the maximum stack depth. Closes #68

### DIFF
--- a/account-integrations/safe/src/SafeWebAuthnPlugin.sol
+++ b/account-integrations/safe/src/SafeWebAuthnPlugin.sol
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 pragma solidity >=0.8.0 <0.9.0;
 
-import "forge-std/Test.sol";
-
 import {BaseAccount} from "account-abstraction/contracts/core/BaseAccount.sol";
 import {IEntryPoint, UserOperation} from "account-abstraction/contracts/interfaces/IEntryPoint.sol";
 
@@ -78,8 +76,6 @@ contract SafeWebAuthnPlugin is BaseAccount {
         UserOperation calldata userOp,
         bytes32 userOpHash
     ) internal override returns (uint256 validationData) {
-        console2.logString("SIG");
-        console2.logBytes(userOp.signature);
         bytes calldata authenticatorData;
         bytes calldata clientData;
         uint256[2] calldata signature;
@@ -91,8 +87,6 @@ contract SafeWebAuthnPlugin is BaseAccount {
             uint i = 0;
             uint dataLen = 32;
             uint256 paramLen = abi.decode(userOp.signature[i:i+dataLen], (uint256));
-            console2.logString("PARAM LEN");
-            console2.logUint(paramLen);
             // Fixed-length params (bytes1, (uint256?), uint256, uint256[2], uint256[2]). Expect 8 slots = 256 bytes
             i += dataLen; // advance index
 
@@ -111,9 +105,6 @@ contract SafeWebAuthnPlugin is BaseAccount {
                 )
             );
             i += dataLen; // advance index
-            console2.logString("FIXED COPY");
-            console2.log(uint8(s.authenticatorDataFlagMask));
-            console2.log("%x", s.clientChallengeDataOffset);
 
 
             bytes calldata calldataLocation;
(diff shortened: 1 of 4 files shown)


account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -67,35 +67,92 @@ contract SafeWebAuthnPlugin is BaseAccount {
...
+        bytes calldata clientData;
+        uint256[2] calldata signature;
+        uint256[2] calldata pubKey;
+        LocalVarStruct memory s;
+
+        {            
+            // parse length of all fixed-length params (including length)
+            uint i = 0;
+            uint dataLen = 32;
+            uint256 paramLen = abi.decode(userOp.signature[i:i+dataLen], (uint256));
```
JohnGuilding, line 89: Out of interest, why do we need to set param length here? I can't see it used again until it is set to get the authenticator data: `paramLen = abi.decode(userOp.signature[i:i + dataLen], (uint256));`

account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -67,35 +67,92 @@ contract SafeWebAuthnPlugin is BaseAccount {
         return _publicKey;
     }
 
+    struct LocalVarStruct {
```
JohnGuilding, line 70: Thoughts on a more descriptive variable name for this? Perhaps something along the lines of:
```solidity
    struct WebAuthnMemoryVars {
        bytes1 authenticatorDataFlagMask;
        uint256 clientChallengeDataOffset;
    }
```

also for readability, thoughts on replacing `s` with something more descriptive like:
```solidity
            webAuthnMemoryVars.authenticatorDataFlagMask,
            , // some number
            webAuthnMemoryVars.clientChallengeDataOffset
```

Open to other name suggestions too

account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -67,35 +67,92 @@ contract SafeWebAuthnPlugin is BaseAccount {
...
+            i += dataLen; // advance index
+
+            // parse length of authenticatorData
+            dataLen = 32;
+            paramLen = abi.decode(userOp.signature[i:i+dataLen], (uint256));
+            i += dataLen; // advance index
+            // assign authenticatorData to sig splice
+            dataLen = paramLen;//((paramLen >> 5) + 1) << 5; // (round up to next slot)
+
+            authenticatorData = userOp.signature[i:i+dataLen];
```
JohnGuilding, line 132: To make sure I understand how this is working:
1. `paramLen` is the length of `authenticatorData` padded to 32 bytes.
2. Once we have the length, we advance the index as we don't care about the length bytes now (with `i += dataLen;`).
3. We then slice the `authenticatorData` from the start of `authenticatorData` to `i+dataLen`, which is the total length of the `authenticatorData` bytes. Thus giving us the correct bytes?

^Please correct me if I'm wrong

One thing I'm not sure on, is why we need to assign `dataLen = paramLen;`, can you explain the reasoning behind that? I played around with the code to just use `paramLength` and the tests still passed for me:

```solidity
        // parse length of authenticatorData
        dataLen = 32;
        paramLen = abi.decode(userOp.signature[i:i + dataLen], (uint256));
        i += dataLen; // advance index
        // assign authenticatorData to sig splice
        authenticatorData = userOp.signature[i:i + paramLen];
        i += ((paramLen >> 5) + 1) << 5; // advance index (round up to next slot)

        // parse length of clientData
        paramLen = abi.decode(userOp.signature[i:i + dataLen], (uint256));
        i += dataLen; // advance index
        // assign clientData to sig splice
        clientData = userOp.signature[i:i + paramLen];
```

account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -67,35 +67,92 @@ contract SafeWebAuthnPlugin is BaseAccount {
...
+            i += ((dataLen >> 5) + 1) << 5; // advance index (round up to next slot)
+
+            // parse length of clientData
+            dataLen = 32;
+            paramLen = abi.decode(userOp.signature[i:i+dataLen], (uint256));
+            i += dataLen; // advance index
+            // assign clientData to sig splice
+            dataLen = paramLen;// ((paramLen >> 5) + 1) << 5; // (round up to next slot)
+            clientData = userOp.signature[i:i+dataLen];
+            i += ((dataLen >> 5) + 1) << 5; // advance index (round up to next slot)
```
JohnGuilding, line 142: Do we need this final advance? The tests still passed for me when I tried removing it

account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -67,35 +67,92 @@ contract SafeWebAuthnPlugin is BaseAccount {
...
-            return 0;
+            i += dataLen; // advance index
+
+
+            bytes calldata calldataLocation;
+            // load fixed length array params (pointers to calldata)
+            dataLen = 2 * 32;
+            calldataLocation = userOp.signature[i:i+dataLen];
+            assembly{
+                signature := calldataLocation.offset
```
JohnGuilding, line 115: To make sure I'm understanding, we assign the offset to the signature variable and not the `calldataLocation` bytes because solidity expects it to point to the location in calldata instead of the whole value?

I'm taking this understanding from this snippet in the solidity docs:
> local variables that refer to statically-sized calldata arrays or calldata structs evaluate to the address of the variable in calldata, not the value itself.

account-integrations/safe/src/lib/FCL_Webauthn.sol
```diff
@@ -76,13 +76,13 @@ library FCL_WebAuthn {
 
     /** @notice Modified from original Fresh Crypto Lib code to use memory instead of calldata */
     function checkSignature(
```
JohnGuilding, line 78: Calldata also needed for `WebAuthn_format` in this file:
```solidity
    function WebAuthn_format(
        bytes calldata authenticatorData,
        bytes1 authenticatorDataFlagMask,
        bytes calldata clientData,
        bytes32 clientChallenge,
        uint256 clientChallengeDataOffset,
        uint256[2] calldata // rs
    ) internal pure returns (bytes32 result)
```

We should remove the new function I added to `FCL_elliptic.sol` at some point too. It's called `ecdsa_verify_memory` and I only I added that to use vars stored in memory.

Then we should call the original function instead in `checkSignature` in this file:
```solidity
// we should change this
bool result = FCL_Elliptic_ZZ.ecdsa_verify_memory(message, rs, Q);

// to this
bool result = FCL_Elliptic_ZZ.ecdsa_verify(message, rs, Q);
``

**Review by JohnGuilding — APPROVED** [review]
Enjoyed reviewing this, took my time and learnt a lot :)

Looks good to me, will merge now as I need to update my PR to build off of this. Left a few comments, let me know what you think!

File: account-integrations/safe/src/lib/FCL_Webauthn.sol (hunk 6 above)
Comment (JohnGuilding) [reply account-integrations/safe/src/lib/FCL_Webauthn.sol:78]: Resolved by https://github.com/getwax/wax/pull/93 which is now merged

File: account-integrations/safe/src/SafeWebAuthnPlugin.sol (hunk 4 above)
Comment (jzaki) [reply account-integrations/safe/src/SafeWebAuthnPlugin.sol:142]: Commented out, had left it in for completeness. If the var is unread before going out of scope, hopefully the compiler would have optimised that out anyway.

File: account-integrations/safe/test/forge/utils/TestHelper.sol [hunk 7]
```diff
@@ -118,5 +118,11 @@ abstract contract TestHelper is Test {
             signature,
             publicKey
         );
+        // console2.logString("authenticatorData");
+        // console2.logBytes(authenticatorData);
+        // console2.logString("flagMask");
+        // console2.logBytes1(authenticatorDataFlagMask);
+        // console2.logString("userOp");
+        // console2.logBytes(userOpSignature);
```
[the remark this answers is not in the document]
jzaki: Removed.

account-integrations/safe/src/SafeWebAuthnPlugin.sol
jzaki: Updated to use the read paramLen (with comments).

account-integrations/safe/src/SafeWebAuthnPlugin.sol
jzaki: I named it for it's purpose of storing multiple local variables as one, not specifically a struct related to webAuthn (although it happens to contain webauthn vars due to the function it's in. Better name given, and comment added.

PR #86, opened 2023-09-18
Linting and cleanup for account-integrations/safe

voltrevo:
I'm finding myself doing some work in here as part of #74. As part of that, I've cleaned up the typing errors and added linting.

Highly recommended to use an IDE that can autofix most linting errors whenever you save the file. Please reach out if you have any trouble with this.

I've manually tested these changes with `yarn hardhat test` using the eth-infinitism bundler.

PR #87, opened 2023-09-18
More cleanup for account-integrations/safe

voltrevo:
# *Dependent PR*

This PR depends on #86. Please merge that PR first.

## About

- Remove ethers-v5 (consolidate on v6 only)
- Improve syncing with bundler using APIs instead of waiting 5 seconds
- `sendUserOpAndWait` sends a user operation and polls for the receipt
- `receiptOf` avoids double await pattern, alleviating the temptation to not wait for the receipt

`yarn hardhat test` is now much faster (not foundry fast, but still).

### DIFF
--- a/account-integrations/safe/package.json
+++ b/account-integrations/safe/package.json
@@ -33,7 +33,6 @@
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-prettier": "^5.0.0",
     "ethers": "^6.4.0",
-    "ethers-v5": "npm:ethers@5.7.2",
     "hardhat": "^2.17.1",
     "hardhat-gas-reporter": "^1.0.8",
     "hardhat-preprocessor": "^0.1.5",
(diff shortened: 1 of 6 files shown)


account-integrations/safe/package.json
```diff
@@ -33,7 +33,6 @@
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-prettier": "^5.0.0",
     "ethers": "^6.4.0",
-    "ethers-v5": "npm:ethers@5.7.2",
```
JohnGuilding, line 36: When I originally copied over these tests and converted them to use ethers v6, I ran into an issue where the bundlerProvider would not work if it wasn't connected to an ethers v5 provider. Hence I added this alias. Did you run into this issue at all? It looks like you just started using ethers v6 for the bundlerProvider with no issues

voltrevo: Oh interesting. Yeah I didn't run into any trouble, wonder what happened.

PR #95, opened 2023-09-18
55 move low level webauthn logic to primitives directory

JohnGuilding:
This PR:
1. Adds a new foundry project inside the primitives directory
2. Moves webauthn contract and accompanying libraries to primitives directory (including test)
3. Symlinks primitives to account-integrations/safe
4. Deletes old files in the safe directory
5. Adds a github workflow for primitives


blakecduncan (approved):
Nice! This lgtm!

primitives/src/WebAuthn.sol
```diff
@@ -3,22 +3,22 @@ pragma solidity ^0.8.12;
...
 import {FCL_WebAuthn} from "./lib/FCL_Webauthn.sol";
 
-contract Webauthn {
+contract WebAuthn {
     function verifySignature(
         bytes calldata authenticatorData,
         bytes1 authenticatorDataFlagMask,
         bytes calldata clientData,
-        bytes32 messageHash,
+        bytes32 clientChallenge,
```
jzaki, line 11: On the cryptographic side (before here) is client challenge always guaranteed to be 32 bytes?

JohnGuilding: Good question, I've seen `bytes32` used elsewhere, but looking at the spec itself, it looks like the challenge should be a minimum of 16 bytes, so it could be different to 32 bytes:
> In order to prevent replay attacks, the challenges MUST contain enough entropy to make guessing them infeasible. Challenges SHOULD therefore be at least 16 bytes long.

https://www.w3.org/TR/webauthn-2/#sctn-cryptographic-challenges

FCL [assumes](https://github.com/rdubois-crypto/FreshCryptoLib/blob/ebf1a485c4e6b9637de2e87e7a8bf1abe16d7b28/solidity/src/FCL_Webauthn.sol#L88) `bytes32` but perhaps this is wrong according to the spec. Added an issue to look into this - https://github.com/getwax/wax/issues/107

jzaki:
Would be good to know how much of `account-integrations/safe/SafeWebAuthnPlugin` can make its way into `primitives/WebAuthn` and remain generic to 4337.
@blakecduncan this could be explored in bringing `WebAuthn` into `account-integrations/kernel` (lower priority than current integration work)

Issue #89, opened 2023-09-18
Fix unauthorized access


voltrevo:
The plugins currently allow `execTransaction` to be called from any address, not just the 4337 entry point.

(Perhaps this should be fixed by another plugin or some other mechanism?)

This is demonstrated by the (currently) failing test added in #88:

```ts
  itif("should not allow execTransaction from unrelated address", async () => {
    const { accountAddress, userWallet, provider } = await setupDeployedAccount(
      ethers.ZeroAddress,
      0,
      "0x",
    );

    const unrelatedWallet = ethers.Wallet.createRandom(provider);

    await receiptOf(
      userWallet.sendTransaction({
        to: unrelatedWallet.address,
        value: 100n * oneEther,
      }),
    );

    const account = SafeECDSAPlugin__factory.connect(
      accountAddress,
      unrelatedWallet, // <-------------------------------------- simply connect with unrelated wallet
    );

    const recipient = ethers.Wallet.createRandom(provider);

    await expect(
      receiptOf(account.execTransaction(recipient.address, oneEther, "0x")),
      //                ~~~~~~~~~~~~~~~ <------------------------ and call execTransaction
    ).to.eventually.rejected; // Should be rejected, but it isn't

    await expect(provider.getBalance(recipient)).to.eventually.equal(0n);
  });
```

voltrevo:
I ran into another stumbling block with my compression plugin. When calling a plugin method msg.sender is the safe, rather than the caller. That's a big problem for checking that the call comes from a trusted source (ie entryPoint or self). As far as I can tell, this would require changes on the safe side.

PR #96, opened 2023-09-18
Safe Recovery Plugin

blakecduncan:


jzaki:
`After the SafeECDSAPlugin is enabled, the original owner of the safe can still execute transactions the normal 'safe' way. Do we want to change this so that when a plugin is enabled you can't still sign transactions the old way?`
I spoke with Andrew about this earlier today, I think a default ECDSA plugin to begin, so original way is not required. Perhaps even an ECDSA multisig plugin that replicats the same functionality. For demos we don't need to remove the original safe multisig functionality, it's a helpful fallback if all validation plugins are removed or become unusable through user config error.

`Since we deploy the SafeECDSAPlugin with the acutal Safe, the plugin doesn't know the address of the safe. So we can't do validation to check msg.sender is the safe.`
There will be a way around this chicken/egg scenario, feel free to create an issue (or not if referring to https://github.com/getwax/wax/issues/89).

`I think it could be a nice UX improvement to make the ECDSA plugin a singleton or registry that anyone can enable. Rather than deploying a plugin for every Safe.`
Yes, also mentioned this morning. Feel free to create an issue :+1:

account-integrations/safe/src/SafeECDSAPlugin.sol
```diff
@@ -71,10 +72,17 @@ contract SafeECDSAPlugin is BaseAccount {
         return _owner;
     }
 
+    function updateOwner(address newOwner) public {
+        // TODO: Check if msg.sender is the Safe. We can't do that now because the Safe
+        // and the plugin are deployed at the same time.  So the plugin doesn't know the Safe address.
+        emit OWNER_UPDATED(_owner, newOwner);
+        _owner = newOwner;
+    }
```
jzaki, line 80: Lets make sure the readme is clear with its state of "pre-alpha" and "not for production use" etc.

account-integrations/safe/src/SafeECDSARecoveryPlugin.sol
```diff
@@ -0,0 +1,56 @@
...
+        address ecdsaPluginAddress,
+        address newValidatingEcdsaAddress
+    ) external onlyStoredEOA {
+        if (safe != storedSafe) {
+            revert ATTEMPTING_RESET_ON_WRONG_SAFE(safe);
+        }
+
+        bytes memory data = abi.encodeWithSignature("updateOwner(address)", newValidatingEcdsaAddress);
+        ISafe(safe).execTransactionFromModule(ecdsaPluginAddress, 0, data, Enum.Operation.Call);
+    }
```
jzaki, line 55: Good start, will have to update and add to this, linking other verifications beyond `onlyStoredEOA`

PR #101, opened 2023-09-22
Make dev faster

voltrevo:
- Set `pollingInterval` to 100ms if chain id is 1337 or 31337 (default = 4000ms)
- Use `dev.period=0` for instant sealing and don't enable mining

This makes `./scripts/start.sh` and `yarn hardhat test` much faster.

Before:

Screen Shot 2023-09-22 at 3 40 29 pm

After:

Screen Shot 2023-09-22 at 3 40 40 pm


jzaki (approved):
Does what it says on the tin.

PR #102, opened 2023-09-22
Clarify webauthn decoding from pr comments

jzaki:
Comments from #85

### DIFF
--- a/account-integrations/safe/src/SafeWebAuthnPlugin.sol
+++ b/account-integrations/safe/src/SafeWebAuthnPlugin.sol
@@ -91,7 +91,7 @@ contract SafeWebAuthnPlugin is BaseAccount {
             uint i = 0;
             uint dataLen = 32;
             uint256 paramLen = abi.decode(userOp.signature[i:i+dataLen], (uint256));
-            // Fixed-length params (bytes1, (uint256?), bytes32, uint256, uint256[2], uint256[2]). Expect 9 slots = 256 bytes
+            // Fixed-length params (bytes1, (uint256?), bytes32, uint256, uint256[2], uint256[2]). Expect 9 slots (288 bytes)
             i += dataLen; // advance index
 
             // decode fixed length params (values to memory)
--- a/account-integrations/safe/src/SafeWebAuthnPlugin.sol
+++ b/account-integrations/safe/src/SafeWebAuthnPlugin.sol
@@ -67,7 +67,9 @@ contract SafeWebAuthnPlugin is BaseAccount {
         return _publicKey;
     }
 
-    struct LocalVarStruct {
+    // Struct declaration to hold multiple local vars.
+    // Prevents stack from getting too deep for evm.
+    struct LocalVarWrapper {
         bytes1 authenticatorDataFlagMask;
         bytes32 clientChallenge;
         uint256 clientChallengeDataOffset;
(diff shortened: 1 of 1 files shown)


account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -92,12 +95,13 @@ contract SafeWebAuthnPlugin is BaseAccount {
             i += dataLen; // advance index
 
             // decode fixed length params (values to memory)
-            dataLen = 4 * 32; //lenFixedParams - 32; // -32 already read length
+            dataLen = paramLen - 32; // length already read
+            dataLen -= 2 * 2 * 32; // exclude fixed length arrays
```
JohnGuilding, line 99: I'm not sure I'm understanding this change, can you explain what's going on here and how this leads to the same value as `dataLen = 4 * 32;`?

account-integrations/safe/src/SafeWebAuthnPlugin.sol
JohnGuilding, line 1: What did you think about [this comment?](https://github.com/getwax/wax/pull/85#issuecomment-1721208197)

JohnGuilding (commented):
Left some questions to clarify a few points :)

account-integrations/safe/src/SafeWebAuthnPlugin.sol
```diff
@@ -130,8 +134,7 @@ contract SafeWebAuthnPlugin is BaseAccount {
             paramLen = abi.decode(userOp.signature[i:i+dataLen], (uint256));
             i += dataLen; // advance index
             // assign authenticatorData to sig splice
-            dataLen = paramLen;//((paramLen >> 5) + 1) << 5; // (round up to next slot)
-
+            dataLen = paramLen;
```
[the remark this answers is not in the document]
jzaki: In that linked suggestion, it moves around how the variables are used throughout. That is:
- `i` is always the index at the start of the data,
- `dataLen` is always used as the length of data to read
- `paramLen` is the length of the parameter

So assigning `dataLen` to be `paramLen` is hopefully helpful towards self-documenting code (and if redundant should be optimised out by the compiler anyway).
Using `paramLen` directly in the slice is different to how the slices are used everywhere else. So even though it would not error, it is less clear as a whole.

account-integrations/safe/src/SafeWebAuthnPlugin.sol
jzaki: The param length is the length of all fixed bytes including the length parameter (you can add this up from the fixed params or print out to see the value).
The first line subtracts the already-read length, the next line subtract two arrays of length 2 containing 32bytes.

account-integrations/safe/src/SafeWebAuthnPlugin.sol
jzaki: I was at a squeeze to get the vars to fit without stack overflow. Adding another call introduces additional var copies and stack usage. I didn't try it for that reason, I can try that as a new PR.

JohnGuilding: Ok makes sense, sounds good to me

account-integrations/safe/src/SafeWebAuthnPlugin.sol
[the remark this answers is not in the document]
JohnGuilding: Thanks for clarifying, that makes sense to me and agree that seems more clear

JohnGuilding (approved):

account-integrations/safe/src/SafeWebAuthnPlugin.sol
JohnGuilding: As discussed in call, clarified changes and updated one of the comments

PR #111, opened 2023-10-08
Add script for starting with hardhat

voltrevo:
Just a variation of `./script/start.sh` which uses hardhat instead of geth.

This requires using the bundler in unsafe mode.

Using hardhat can be advantageous for the different diagnostics it provides, especially integration with `console.log`.

### DIFF
--- a/account-integrations/safe/script/start-hardhat.sh
+++ b/account-integrations/safe/script/start-hardhat.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+set -meuo pipefail
+
+function cleanup {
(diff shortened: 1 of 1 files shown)


account-integrations/safe/script/start-hardhat.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
jzaki, line 2: Comment benefits of use. (Also helpful in start.sh and readme)

jzaki:
Would be good to document benefits of both ways of testing.

Something like:

- `start-hardhat.sh` used for different diagnostics including console.log, but no checks of bundler restrictions
- `start.sh` used for testing against constraints of the bundler specification.

jzaki (approved):

voltrevo:
Thanks @jzaki. Docs followup: https://github.com/getwax/wax/pull/115.

PR #114, opened 2023-10-09
Small updates to enable a testnet build for deployment

blakecduncan:
Main difference is I created a testnet config file and a testnet setup file.  Not much difference right now but figured breaking them out into their own files was a good idea.  Also, added an RPC env variable to the testnet config.

I have this branch deployed here:
https://test-inpage-build.d2q4o4f8gnoq5j.amplifyapp.com/

We will need to figure out what domain we want for this.


JohnGuilding (approved):
LGTM!

JohnGuilding:
@blakecduncan I get an insufficient funds error when clicking "Add funds" for a SimpleAccount in the demo. Does that mean funds have run out somewhere, or that I need to fund an account myself?

blakecduncan:
> @blakecduncan I get an insufficient funds error when clicking "Add funds" for a SimpleAccount in the demo. Does that mean funds have run out somewhere, or that I need to fund an account myself?

Yeah it won't work with the default seed phrase that is in the popup.  You'll need to input a seed phrase that has funds when the popup shows up.  I have one if you want to test it out.

I didn't end up making any changes to the default account that is used (this is the hardhat account which is why it doesn't work on testnet).  I was curious to get some opinions from the team. I was thinking we have two options.

1. Keep the demo as is.  And maybe we can include a funded seed phrase on the flier we hand out at Devconnect so people can test it out.
2. I can change the seed phrase in the code so it auto-populates with a funded address.

Number two is probably the easiest user flow for Devconnect.  I kind of liked the idea of #1 because it made it so the user had to input a seed phrase (even if we give it to them).  Was thinking this might make the wallet less likely to be drained.  But it does add friction to the demo which we don't want.

blakecduncan:
> t

Yeah I'll get a PR up to use option 2.  I'll use an env variable so that it's not in the code and makes it a little less likely to get drained.

This is a good explanation for the longer term.  I agree that a paymaster with webauthn would probably be the best demo.  I honestly think you hit the mark with your description and I don't think it would be too much work to do the paymaster approach.

Issue #119, opened 2023-10-11
Use compression: use compression safe plugin in in-page wallet


voltrevo:

voltrevo:
PR: https://github.com/getwax/wax/pull/123

PR #120, opened 2023-10-11
Add compression plugin

### LINKED ISSUE #113: Use compression: safe plugin
This is for encoding the `calldata` field of the `UserOperation`. The wallet needs to decode this itself.

voltrevo:
Resolves #113.

(#113 would originally have included applying this in the in-page wallet as well, but I made that a separate issue that can build on top of this work: #119).

### DIFF
--- a/account-integrations/safe/test/hardhat/integration/SafeCompressionPlugin.test.ts
+++ b/account-integrations/safe/test/hardhat/integration/SafeCompressionPlugin.test.ts
@@ -1,0 +1,198 @@
+import { expect } from "chai";
+import { getBytes, resolveProperties, ethers } from "ethers";
+import { UserOperationStruct } from "@account-abstraction/contracts";
+import { getUserOpHash } from "@account-abstraction/utils";
+import {
+  AddressRegistry__factory,
+  FallbackDecompressor__factory,
+  SafeCompressionFactory__factory,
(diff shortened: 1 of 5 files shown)


JohnGuilding (approved):
LGTM!

account-integrations/safe/test/hardhat/integration/SafeCompressionPlugin.test.ts
```diff
@@ -0,0 +1,198 @@
...
+          data: "0x",
+        },
+      ],
+      [],
+    );
+
+    const userOpCallData = compressionAccount.interface.encodeFunctionData(
+      "decompressAndPerform",
+      [compressedActions],
+    );
```
jzaki, line 145: Current implementation requires dApps (wallets or SDKs) to be compression-aware and wrap users' functions in `decompressAndPerform`. For broader use, this can eventually be implemented in an 'aggregator'?

voltrevo: I think the wallet will always need to be compression-aware, similar to how it needs to be 4337-aware. Compression means that you actually sign a different message, so you can't give control over what message is signed to a third party. However, the same is true for using 4337 itself, so I'm not sure I understand your concern.

Issue #112, opened 2023-10-09
Use compression: top-level in-page


voltrevo:
This is the compression applied to the EOA transaction (the 'top-level') which encodes the call to the entry point.

This issue is for demonstrating this within the in-page demo (bundler simulation).

PR #123, opened 2023-10-12
Use account-level compression in in-page demo

### LINKED ISSUE #119: Use compression: use compression safe plugin in in-page wallet

voltrevo:
# *Dependent PR*

This PR depends on #120, please merge that PR first.

## Description

Resolves #119.

Adds `SafeCompressionAccount` to in-page wallet:

Screen Shot 2023-10-16 at 15 11 25

Adds dApp for sending eth:

Screen Shot 2023-10-16 at 15 13 51

New logs with prefix `logBytes` for logging about how many bytes are involved in different things:

Screen Shot 2023-10-16 at 15 31 06

Here we can see that 25 bytes are needed by `decompressAndPerform` to tell it to send 0.1 ETH to the specified address (in future reducible to 8 bytes if the recipient is the in the address registry). Note that this doesn't include specifying the gas limit and other fields of the `userOp`, those are not part of account-level compression and will instead be included in top-level compression (#112).

This still requires 100 bytes of calldata into the account to encode the `decompressAndPerform` call as a whole (4 bytes to specify method, 32 bytes to say where the `stream` argument is, 32 bytes to specify the length of `stream`, 25 bytes for the actual data, and 7 padding bytes). These 75 bytes overhead can be reduced to 1 byte (maybe less) using #112.

Zooming out to the actual EOA transaction (`entryPoint.handleOps`), the transaction requires 952 bytes in total (71% zeros). This will be reduced dramatically later by #112.

For comparison, here are the numbers when doing the same transaction from `SafeECDSAAccount`:

Screen Shot 2023-10-16 at 15 41 12

### Other changes

- Import new contracts (no other/real contract changes)
- New `waxInPage` internal API for getting the account only if it exists
- Change `eth_estimateGas` when `from` is the user's account. By default, the node will estimate the cost of performing that call from that address. This change will instead estimate the entry point making the appropriate call to the user's account to *then* make the specified call.
  - This solves gas issues with compression since it can make a big difference to include the work the account does to decode the action before doing it
  - Also replaces previous logic to add gas when sending eth to a new account
- Implement `eth_getTransactionReceipt` (needed by ethers when sending eth (wasn't needed when sending data for some reason))

### DIFF
--- a/demos/inpage/src/EthereumApi.tsx
+++ b/demos/inpage/src/EthereumApi.tsx
@@ -9,7 +9,6 @@ import IBundler from './bundlers/IBundler';
 import waxPrivate from './waxPrivate';
 import ethereumRequest from './ethereumRequest';
 import calculateUserOpHash from './helpers/calculateUserOpHash';
-import { hexLen } from './helpers/encodeUtils';
 
 // We need a UserOperation in order to estimate the gas fields of a
 // UserOperation, so we use these values as placeholders.
@@ -19,8 +18,6 @@ const temporarySignature = [
   '123456fe2807660c417ca1a38760342fa70135fcab89a8c7c879a77da8ce7a0b5a3805735e',
   '95170906b11c6f30dcc74e463e1e6990c68a3998a7271b728b123456',
 ].join('');
-
-const extraGasForTransferToNewAddress = 20_000n;
 
 type StrictUserOperation = {
   sender: string;
@@ -200,7 +197,7 @@ export default class EthereumApi {
         });
       }
 
-      const account = await this.#waxInPage._getAccount(waxPrivate);
+      const account = await this.#waxInPage._getOrCreateAccount(waxPrivate);
 
       connectedAccounts = [account.address];
 
@@ -213,7 +210,7 @@ export default class EthereumApi {
       await this.#waxInPage.storage.connectedAccounts.get(),
 
     eth_sendTransaction: async (...txs) => {
-      const account = await this.#waxInPage._getAccount(waxPrivate);
+      const account = await this.#waxInPage._getOrCreateAccount(waxPrivate);
       const contracts = await this.#waxInPage.getContracts();
 
       const sender = txs[0].from ?? account.address;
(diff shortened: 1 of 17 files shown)


voltrevo:
Resolves #119

demos/inpage/src/WaxInPage.tsx
```diff
@@ -161,6 +173,11 @@ export default class WaxInPage {
 
     const assumedEntryPoint = viewer.connectAssume(EntryPoint__factory, []);
 
+    const assumedAddressRegistry = viewer.connectAssume(
```
JohnGuilding, line 176: nitpick - is there a reason `viewer.connectAssume` is used for entrypoint and address registry outside of the `contracts` object? Rather than when creating the object like the contracts. Seems arbitrary but wasn't sure if there was a reason

demos/inpage/src/EthereumApi.tsx
```diff
@@ -286,18 +284,6 @@ export default class EthereumApi {
...
-
-            if (BigInt(value ?? 0) !== 0n) {
-              const recipientBalance =
-                await this.#waxInPage.ethersProvider.getBalance(tx.to);
-
-              const txCount =
-                await this.#waxInPage.ethersProvider.getTransactionCount(tx.to);
-
-              if (recipientBalance === 0n && txCount === 0) {
-                gas += extraGasForTransferToNewAddress;
```
JohnGuilding, line 298: So to clarify, you updated `eth_estimateGas` to estimate gas as an entry point transaction when `from` is a users account, and that solved this issue?

JohnGuilding (approved):
Looks good. Looks like there are two small merge conflicts

demos/inpage/src/EthereumApi.tsx
voltrevo: Yep

demos/inpage/src/WaxInPage.tsx
voltrevo: Yeah it's so that we can pass in the address as an argument to the other contracts.

demos/inpage/src/helpers/encodeUtils.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (x === 0n) {
+      break;
+    }
+  }
+
+  let result = '0x';
+
+  for (let i = 0; i < segments.length; i++) {
+    const keepGoing = i !== segments.length - 1;
```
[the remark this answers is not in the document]
voltrevo: Nah I copied them from `bls-wallet`.

Not sure what you mean by whether `keepGoing` is conceptually `lastIndex`. If you're referring to [`lastIndexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf), then no. If you just mean whether `keepGoing` is a boolean that says whether `i` is the last index, then yes.

Conceptually, `keepGoing` makes a bit more sense with the decoder in mind - each byte uses the high bit as a flag for the decoder to indicate whether it should keep going. This allows us to avoid using a dedicated byte (or more) to encode the length, allowing for 0-127 to be encoded in a single byte.

PR #125, opened 2023-10-17
75 demonstrate recovery module in in page wallet

JohnGuilding:
This PR adds a basic recovery flow to the inpage. This includes setting up the recovery module and also recovering the account via the module.

There are a few improvements to be made to this flow, notably:
- Remove use of two seed phrases for recovery - not an ideal flow but was a quick way to get the demo working in it's current form
- Send txs via bundler (as opposed to calling contracts the non-4337 way)
- Deploy recovery plugin on initial contracts deploy (can't do this yet as tightly coupled to a single account rather than being a singleton that any account can use)
- The pattern so far has been a ts wrapper class per module. The recovery module is a little different as it is currently not a standalone module that executes txs. So we should explore how we want to manage this as a recovery wrapper wouldn't match the `IAccount` interface. I've tightly coupled the recovery functions to the ECDSA wrapper for now but maybe we want to do something like dependency injection to compose the wrapper class.
- Prove owner of recovery address is in possession of private key, and prevent replay of recovery address proof e.g. nonce. Also other checks on the contracts side like zero address checks

^Going to be working on above issues next

**Recovery page on page load**

Screenshot 2023-10-18 at 11 48 05

**Recovery page after recovering an account**

Screenshot 2023-10-18 at 11 47 41

If interested in testing it out, you can test it with these values:
- New recovery address: `0x803EaF4e5FBf36B0ddE3E3104d5a382Eb0238866`
- New owner account seed phrase: `expect discover swamp appear divide barely silly hazard turtle bubble level scatter`
- Recovery account seed phrase: `claim garden try absorb alert decorate rural rival hunt mosquito gift course`

### DIFF
--- a/demos/inpage/demo/LinksPage.tsx
+++ b/demos/inpage/demo/LinksPage.tsx
@@ -5,6 +5,7 @@ import usePath from './usePath';
 import DemoContext from './DemoContext';
 import runAsync from './helpers/runAsync';
 import config from './config/config';
+import waxPrivate from '../src/waxPrivate';
 
 const addFundsDefault = (() => {
   if (config.rpcUrl === 'http://127.0.0.1:8545') {
@@ -33,9 +34,22 @@ const LinksPage = () => {
             'fund-new-account',
           );
 
+          const account = await demo.waxInPage._getAccount(waxPrivate);
+          const { ownerAddress } = account.toData();
+
           await (
             await admin.sendTransaction({
               to: address,
+              value: ethers.parseEther(
+                config.addFundsEthAmount ?? addFundsDefault,
+              ),
+            })
+          ).wait();
+
+          // TODO: If default config, do this
+          await (
+            await admin.sendTransaction({
+              to: ownerAddress,
               value: ethers.parseEther(
                 config.addFundsEthAmount ?? addFundsDefault,
               ),
@@ -52,6 +66,14 @@ const LinksPage = () => {
       </Button>
       <Button secondary onPress={() => setPath('/greeter')}>
         Greeter dApp
+      </Button>
+      <Button
+        secondary
+        onPress={() => {
+          setPath('/recovery');
+        }}
+      >
+        Recovery Example
       </Button>
       <Button
         secondary
--- a/demos/inpage/demo/LinksPage.tsx
+++ b/demos/inpage/demo/LinksPage.tsx
@@ -5,6 +5,7 @@ import usePath from './usePath';
 import DemoContext from './DemoContext';
 import runAsync from './helpers/runAsync';
 import config from './config/config';
+import SafeECDSAAccountWrapper from '../src/accounts/SafeECDSAAccountWrapper';
 
 const addFundsDefault = (() => {
   if (config.rpcUrl === 'http://127.0.0.1:8545') {
@@ -65,14 +66,16 @@ const LinksPage = () => {
       <Button secondary onPress={() => setPath('/greeter')}>
         Greeter dApp
       </Button>
-      <Button
-        secondary
-        onPress={() => {
-          setPath('/recovery');
-        }}
-      >
-        Recovery Example
-      </Button>
+      {account instanceof SafeECDSAAccountWrapper && (
+        <Button
+          secondary
+          onPress={() => {
+            setPath('/recovery');
+          }}
+        >
+          Recovery Example
+        </Button>
+      )}
       <Button
         secondary
         onPress={async () => {
--- a/demos/inpage/demo/LinksPage.tsx
+++ b/demos/inpage/demo/LinksPage.tsx
@@ -5,7 +5,6 @@ import usePath from './usePath';
 import DemoContext from './DemoContext';
 import runAsync from './helpers/runAsync';
 import config from './config/config';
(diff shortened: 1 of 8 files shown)


demos/inpage/demo/LinksPage.tsx
```diff
@@ -16,26 +16,38 @@ const addFundsDefault = (() => {
...
+          await (
+            await admin.sendTransaction({
+              to: account.address,
+              value: ethers.parseEther(
+                config.addFundsEthAmount ?? addFundsDefault,
+              ),
+            })
+          ).wait();
+
+          // TODO: (merge-ok) ensure wallet is already funded
```
JohnGuilding, line 47: we also need to fund the owner of the wallet as the owner needs to call "enable module" and pay for that transaction

demos/inpage/src/accounts/SafeECDSAAccountWrapper.ts
```diff
@@ -135,4 +148,114 @@ export default class SafeECDSAAccountWrapper implements IAccount {
...
+    await (
+      await owner.sendTransaction({
+        to: recoveryWallet.address,
+        value: ethers.parseEther('0.1'),
+      })
+    ).wait();
+
+    const pluginAddress = await plugin.myAddress();
+    await recoveryPlugin
+      .connect(recoveryWallet)
```
JohnGuilding, line 246: We need the recovery account seed phrase or private key as we need to execute `resetEcdsaAddress` as the recovery account

demos/inpage/src/accounts/SafeECDSAAccountWrapper.ts
```diff
@@ -135,4 +148,114 @@ export default class SafeECDSAAccountWrapper implements IAccount {
...
+      moduleEnabled = await safeProxy.isModuleEnabled(recoveryPluginAddress);
+      assert(moduleEnabled, 'module not enabled');
+
+      this.recoveryAddress = recoveryAddress;
+      await this.waxInPage.storage.accounts.set([this.toData()]);
+    }
+  }
+
+  async recoveryAccount(
+    newOwnerAccountSeedPhrase: string,
```
blakecduncan, line 207: Hmm I didn't think about this.  Curious to here your thoughts about whether we should pass the seed phrase here or the address.  I like the idea of the recovery account being able to rest the address without knowing the seed phrase.  However, it might make it more confusing for the demo to enter an address since the Safe still has the same address.

blakecduncan: It might not be that confusing if we explain/demonstrate better that since right now we are using an ECDSAplugin, an eoa can actually control the safe.  So recovery is really resetting the owning EOA

demos/inpage/demo/Recovery.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+      {recoverySet && (
+        <section className="recovery-section">
+          <Heading>Recover account</Heading>
+          <div>
+            New owner account seed phrase:{' '}
+            <input
+              type="text"
+              onInput={(e) =>
+                setNewOwnerAccountSeedPhrase(e.currentTarget.value)
```
blakecduncan, line 94: I wonder if we should allow this to pass a seed phrase or a PK.  I'm just thinking it's easier to get several private keys from metamask

blakecduncan (approved):
Nice!  Just added a few thoughts but this looks really good!

Also, I just noticed.  Recovery doesn't work with the simple account, I wonder if there is an easy way to hide the recovery button if you select that option

JohnGuilding:
> Also, I just noticed. Recovery doesn't work with the simple account, I wonder if there is an easy way to hide the recovery button if you select that option

Yeah I did add the check on the recovery page instead (without throwing an error or message so UX improvement would be needed there anyway), as thought we may want to add a recovery option for the simple account too since it is upgradeable. But I think hiding the recovery option until if/when we add that makes more sense - will update

demos/inpage/demo/Recovery.tsx
JohnGuilding: Yeah that's a good point regarding ease of access. If we can't solve that by devconnect then think swapping to a private key makes more sense. I'm gunna be thinking about how we can minimise the need to pass any private key or seed phrase for the demo, so hopefully we can do that by devconnect

demos/inpage/src/accounts/SafeECDSAAccountWrapper.ts
JohnGuilding: Yeah I originally just passed the new owner address in as that's all the contract call needs. The issue is that broke the wrapper API as it has a private key field, so there would be a mismatch between the address and private key in the demo the way things are. Agree passing the address is easier/better.

Perhaps we can do some brainstorming and jump on a call to discuss better ways of doing this?

JohnGuilding: Regarding the confusion with the address, yeah feel like we should think about what address info we're showing the user and what makes more sense. We don't want them to give them like 4 addresses to remember but the delineation between safe address and owner sounds reasonable

PR #126, opened 2023-10-18
Allow passing a default deployer seed phrase

blakecduncan:
Right now when loading [the deployed demo](https://main.d2q4o4f8gnoq5j.amplifyapp.com/) the user needs to put in a funded wallet so the contracts can be deployed.  This change allows us to see a default deployer wallet that we can fund.  This will make the UX a little more streamlined for dev connect.  I also use an ENV variable so we don't have to commit the seed phrase.

For more context see this convo: https://github.com/getwax/wax/pull/114#issuecomment-1759471539

### DIFF
--- a/demos/inpage/demo/config/config.testnet.ts
+++ b/demos/inpage/demo/config/config.testnet.ts
@@ -16,7 +16,7 @@ const config: ConfigType = {
     'https://sepolia-rollup.arbitrum.io/rpc',
 
   // If a default seed phrase is not provided, a user must provide their own.
-  deployerSeedPhrase: (import.meta.env.VITE_RPC_URL as string) ?? '',
+  deployerSeedPhrase: (import.meta.env.DEPLOYER_SEED_PHRASE as string) ?? '',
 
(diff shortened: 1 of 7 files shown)


JohnGuilding (approved):
Looks good!

demos/inpage/demo/config/config.testnet.ts
```diff
@@ -15,6 +15,9 @@ const config: ConfigType = {
     (import.meta.env.VITE_RPC_URL as string) ??
     'https://sepolia-rollup.arbitrum.io/rpc',
 
+  // If a default seed phrase is not provided, a user must provide their own.
+  deployerSeedPhrase: (import.meta.env.VITE_RPC_URL as string) ?? '',
```
JohnGuilding, line 19: Is `VITE_RPC_URL` the right env var for the seed phrase?

blakecduncan: Good catch, I changed it in my config locally but forgot to change it in the template config

PR #127, opened 2023-10-19
Add skeleton for the sdk in the monorepo

blakecduncan:
Not doing anything here other than just adding the project and making sure it builds for `esm` and `cjs` outputs.  Also, just wanted to make sure we are cool with where I put it in the mono repo


account-integrations/sdk/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@getwax/sdk",
```
blakecduncan, line 2: Thought this could be a good name to deploy it to

account-integrations/sdk/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "./dist/cjs/index.js",
+  "module": "./diist/esm/index.js",
```
JohnGuilding, line 6: Is `diist` a typo?

account-integrations/sdk/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@getwax/sdk",
+  "version": "1.0.0",
```
JohnGuilding, line 3: Don't know if this matters too much but since this is early doors on the sdk, would we consider starting on a version number like `0.1.0` to indicate it's development stage?

account-integrations/sdk/package.json
JohnGuilding: Like the name 👍

JohnGuilding (approved):
Skeleton looks good - left two questions

account-integrations/sdk/package.json
blakecduncan: yeah, good call.  I'll actually change it to 0.0.1

PR #130, opened 2023-10-20
Update recovery plugin to be singleton

JohnGuilding:
This PR:
1. Updates `SafeECDSARecoveryPlugin` so that it can be used by multiple accounts and is not tightly coupled to a single `SafeECDSAPlugin`.
2. Updates inpage demo to use updated contracts
3. Updates inpage demo to take private keys instead of seed phrases. Private keys are easier to source when running the demo (when using funded hardhat accounts), compared to seed phrases where you have the extra step of funding them first. This change is just to make running through the demo easier as we can use the logged hardhat accounts. The idea is to remove the need to pass private keys or seed phrases in at all, so this is a temporary solution anyway
4. Adds a hardhat task for funding accounts
5. Adds forge tests for new contract changes

### DIFF
--- a/account-integrations/safe/src/SafeECDSARecoveryPlugin.sol
+++ b/account-integrations/safe/src/SafeECDSARecoveryPlugin.sol
@@ -44,6 +44,12 @@ contract SafeECDSARecoveryPlugin {
 
     constructor() {}
 
+    function getEcdsaRecoveryStorage(
+        address owner
+    ) external view returns (ECDSARecoveryStorage memory) {
+        return ecdsaRecoveryStorage[owner];
+    }
+
     modifier onlyRecoveryAccount(address currentOwner) {
         address recoveryAccount = ecdsaRecoveryStorage[currentOwner]
             .recoveryAccount;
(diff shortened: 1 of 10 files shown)


account-integrations/safe/src/SafeECDSARecoveryPlugin.sol
```diff
@@ -22,35 +22,81 @@ interface ISafe {
     ) external returns (bool success);
 }
 
+interface ISafeECDSAPlugin {
+    function getOwner(address safe) external view returns (address);
+}
+
+struct ECDSARecoveryStorage {
+    address recoveryAccount;
```
blakecduncan, line 30: Consideration for the future: I'm starting to really like the term `Guardian` over `recoveryAccount`. I remember I really struggled to think of a good word for this with the BLS recovery.  It also seems like a lot of people in the ecosystem use Guardian as well.

blakecduncan (approved):
Nice, i think this is a really good improvement 💯

account-integrations/safe/src/SafeECDSARecoveryPlugin.sol
JohnGuilding: Ditto - I like that term too. Game to transition to it

PR #135, opened 2023-10-26
131 recover account via signature

JohnGuilding:
This PR:
- Improves guardian privacy by storing a hash instead of a guardian address
- Creates a `RECOVERY_HASH_DOMAIN` to improve recovery security
- Stores recovery hashes by safe address instead of owner address. This is because it is possible that a single owner can manage more than one safe. So there would be a collision when that owner enabled a guardian with their second safe.
- Passes a signature signed by the new owner to `resetEcdsaAddress` to prove recovery is accepted by the new owner
- Adds a check to ensure the module is enabled before it can be used


jacque006 (approved):
@jzaki may want to do a pass as well

PR #139, opened 2023-10-30
136 run all integration tests against bundler

JohnGuilding:
**What has this PR changed:**
- Assumes `addRecoveryAccount` is called from a safe, this is the assumption on how the owner would normally call this function through a userOp. Maybe we also want to cater to the assumption that `addRecoveryAccount` could be called directly by the owner (see comment).
- Runs all integration tests through a bundler 
  - Bls plugin and recovery plugin didn't do this before 
- Shares `setupTests` logic in dedicated file
- Uses an `admin` account for actions such as deploying contracts
- Wraps admin account in an ethers `NonceManager` to prevent several `nonce too low` errors.
  - Also derives second account from mnemonic to avoid clashes with hardhat account 0 which is default account the eth-infinitism bundler uses
- Moves gas estimation logic to dedicated function - `getGasEstimates()`
- Moves logic to create user ops to dedicated file - `createUserOp.ts`

**How to test these changes:**
1. Check env vars are correct according to `.env.example`
2. run `./script.start.sh` in `account-integrations/safe`
3. run `yarn run bundler` in the eth infinitism bundler at commit `1b154c9`
4. Run `yarn hardhat test` in `account-integrations/safe`


account-integrations/safe/src/SafeECDSARecoveryPlugin.sol
```diff
@@ -71,25 +70,24 @@ contract SafeECDSARecoveryPlugin {
 
     function addRecoveryAccount(
         bytes32 recoveryHash,
-        address safe,
+        address owner,
         address ecsdaPlugin
     ) external {
-        if (safe == address(0)) revert SAFE_ZERO_ADDRESS();
+        address safe = msg.sender;
```
JohnGuilding, line 76: @jzaki do you think we should change this to assume it can be called elsewhere beyond a safe, e.g. directly by the owner?

To do that we would need to change how we use `msg.sender`, as it would be different in each case. If calling directly, we couldn't assume `msg.sender` is the safe. Maybe we could do something like the following in `addRecoveryAccountWithoutMsgSender`:

```solidity
    // How it's currently done
    function addRecoveryAccount(
        bytes32 recoveryHash,
        address owner,
        address ecsdaPlugin
    ) external {
        address safe = msg.sender;

        bool moduleEnabled = ISafe(safe).isModuleEnabled(address(this));
        if (!moduleEnabled) revert MODULE_NOT_ENABLED();

        address expectedOwner = ISafeECDSAPlugin(ecsdaPlugin).getOwner(safe);
        if (owner != expectedOwner) revert INVALID_OWNER(expectedOwner, owner);

        ecdsaRecoveryStorage[safe] = ECDSARecoveryStorage(recoveryHash);
    }

    // How we could change it to not rely on msg.sender (pseudocode)
    function addRecoveryAccountWithoutMsgSender(
        bytes32 recoveryHash,
        bytes memory ownerSignature,
        address safe,
        address ecsdaPlugin
    ) external {
        bool moduleEnabled = ISafe(safe).isModuleEnabled(address(this));
        if (!moduleEnabled) revert MODULE_NOT_ENABLED();

        bytes32 ownersSafeHash = keccak256(abi.encodePacked(safe));
        bytes32 ethSignedMessage = ownersSafeHash.toEthSignedMessageHash();
        address owner = ethSignedMessage.recover(ownerSignature);

        address expectedOwner = ISafeECDSAPlugin(ecsdaPlugin).getOwner(safe);
        if (owner != expectedOwner) revert INVALID_OWNER(expectedOwner, owner);

        ecdsaRecoveryStorage[safe] = ECDSARecoveryStorage(recoveryHash);
    }
```

maybe another option might be to create functions for each scenario like `addRecoveryAccountFromSafe` and `addRecoveryAccountFromOwner`. Feels better to use one function though.

Curious to get your thoughts

account-integrations/safe/test/hardhat/utils/createUserOp.ts
```diff
@@ -0,0 +1,169 @@
...
+    factoryAddress,
+    safeProxyFactory.interface.encodeFunctionData("createProxyWithNonce", [
+      singletonAddress,
+      encodedInitializer,
+      73,
+    ]),
+  ]);
+
+  return { initCode, deployedAddress };
+};
```
jacque006, line 73: Is there a reason you went this route vs. using the contract factory pattern that is used for non-recovery plugins?

jacque006 (approved):
Think you just need feedback from James and then you can :shipit: . I will update my inflight PR if this goes in first.

account-integrations/safe/test/hardhat/utils/createUserOp.ts
JohnGuilding: The reason this route is done is because the original safe 4337 [example test ](https://github.com/safe-global/safe-contracts/blob/main/test/integration/Safe.ERC4337.spec.ts) does something like this. So these plugins haven't been updated to use a factory

At the moment there is no shared way that we're creating these plugins, some use a dedicated factory and others use this method.

We probably need to decide how we want to deploy these plugins. We could treat them as 4337 accounts - there is a pattern of using a contract factory for deploying 4337 accounts (e.g SimpleAccount). Although if the plugins are singletons (which some of them are), then deploying them from a factory might not be appropriate

jacque006 (approved):

PR #141, opened 2023-11-03
Adding the accounts code to the sdk

blakecduncan:
I thought a good first thing to do would be to bring the accounts into the SDK.  I had to do some refactoring because this code doesn't have access to WAX inpage.  

Some notes:
- I didn't know how to set up the sym link that @voltrevo did in the inpage directory with `hardhat`. I think we may want to do that in the SDK directory as well so that we can build the contracts there.
- For some of the functions I pass an `admin` account so we can deploy the contracts if they aren't deployed (see comments in the code).
- I think the recovery logic needs to be updated with @JohnGuilding new changes for the contracts
- Also, if we want to keep this name, we need to create a namespace in npm for `@getwax` I can't deploy it under this name until we do that.

Also, shout out to Andrew for making these account wrappers.  I just edited the existing code in the demo so that it can be used in an sdk.


account-integrations/sdk/src/accounts/SafeECDSAAccontWrapper.ts
```diff
@@ -0,0 +1,249 @@
+import { ethers } from 'ethers';
+import { z } from 'zod';
+import {
+  SafeECDSAPlugin,
+  SafeECDSAPlugin__factory,
+  SafeECDSARecoveryPlugin,
+  Safe__factory,
+  SafeECDSAFactory,
+} from '../../../../demos/inpage/hardhat/typechain-types';
```
blakecduncan, line 9: We will probably want to create a hardhat folder in the `sdk` directory so we don't have to link to the demo

account-integrations/sdk/src/accounts/SafeECDSAAccontWrapper.ts
```diff
@@ -0,0 +1,249 @@
...
+      privateKey: this.privateKey,
+      ownerAddress: this.ownerAddress,
+      recoveryAddress: this.recoveryAddress,
+    };
+  }
+
+  static async createRandom(
+    // TODO: clean up parameters needed for getContracts
+    provider: ethers.BrowserProvider,
+    admin: ethers.Signer
```
blakecduncan, line 64: Added these TODOs because I wasn't sure if we wanted to have to pass an admin signer to this function.  However, it's needed right now in case we need to deploy the contracts in the `getContracts()` function

account-integrations/sdk/src/accounts/SafeECDSAAccontWrapper.ts
```diff
@@ -0,0 +1,249 @@
...
+  async sign(
+    _userOp: EthereumRpc.UserOperation,
+    userOpHash: string,
+  ): Promise<string> {
+    const ownerWallet = new ethers.Wallet(this.privateKey);
+
+    return await ownerWallet.signMessage(ethers.getBytes(userOpHash));
+  }
+
+  // TODO: update this with the new recovery module
```
blakecduncan, line 148: Didn't get a chance to update this logic with the recovery plugin changes.  We will need to do that

jacque006 (approved):
@voltrevo @JohnGuilding I'm ok for this to go in as-is, and we can address TODOs as we integrate into InPage demo & other usage points. Thoughts?

voltrevo:
@jacque006 Yup makes sense. I'll take a look at that symlink now.

PR #158, opened 2023-12-06
Use top-level compression

### LINKED ISSUE #112: Use compression: top-level in-page

voltrevo:
## What is this PR doing?

Adds top-level compression to the inpage demo. This means the internal bundler calls `HandleOpsCaller` instead of the `EntryPoint`, so that the top-level calldata can be compressed off-chain, and `HandleOpsCaller` decompresses it before passing it on to `EntryPoint`.

Send a transaction using the inpage demo. In the console you should see a log like `logBytes: EntryPoint tx is 319 bytes: ...`.

Top-level compression can be turned off using `waxInPage.setConfig({ useTopLevelCompression: false })` (in the console).

If you send another transaction, you should see a log like `logBytes: EntryPoint tx is 951 bytes: ...`.

Depending on the account type used and the type of transactions, the numbers will be different, but it should be dramatically fewer bytes when compression is enabled.

Resolves #112.

## Checklist

## Guidelines

### DIFF
--- a/account-integrations/compression/README.md
+++ b/account-integrations/compression/README.md
@@ -44,7 +44,7 @@ paragraph more thoroughly and document it.)
 
 ## Requirements
 
-- The bundler needs to use [`EntryPointCaller`](./src/EntryPointCaller.sol)
+- The bundler needs to use [`HandleAggregatedOpsCaller`](./src/HandleAggregatedOpsCaller.sol)
   (or something better) (including the off-chain compression needed to call it
   correctly)
 - The account (aka SCW) needs to decompress its `userOp.calldata` field similar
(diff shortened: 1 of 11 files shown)


voltrevo:
TODO: On-chain encoding

jacque006 (approved):
Some notes/followup from offline call w/ @voltrevo :
- May be worth effort as a team to start using more [natspec](https://docs.soliditylang.org/en/latest/natspec-format.html) documentation on our contracts/public/external functions to make it easier for people to understand & integrate
- End to end test with a bundler modified with compression, similar to current safe tests
- Instead of using a config switch for compression, instead simulate both a compressed and non-compressed UserOp and choose the one that is cheapest. Show gas savings.
- Might be beneficial to have the compression logic in Solidity as well, bundlers could call?

All future ideas/work out of scope of this.

PR #169, opened 2023-12-08
Use AddressRegistry in inpage demo

### LINKED ISSUE #122: Compression: Use address registry in in-page demo
- [ ] Add UI for registering addresses
- [ ] Implement registry lookup when encoding actions (see comment referencing `#122` in `SafeCompressionAccountWrapper.ts`)

This allows using 3 bytes instead of 20 when encoding addresses.

voltrevo:
# *Dependent PR*

This PR depends on https://github.com/getwax/wax/pull/165. Please merge that PR first.

## What is this PR doing?

- Makes use of the AddressRegistry in the inpage demo
- Makes `Add Funds` primary when the account has no funds
- Fixed bug in `encodeRegIndex`

Demo:
[Demo video](https://www.youtube.com/watch?v=En-i1brvXss)

Btw I forgot to mention that in a real product/deployment you'd probably proactively register the user's address without them doing it explicitly (and other common addresses). Another one of those things where we're showcasing dev tools instead of making a product. Curious to sync on that idea if anyone thinks differently.

Resolves #122.

## Checklist

## Guidelines

### DIFF
--- a/demos/inpage/demo/LinksPage.tsx
+++ b/demos/inpage/demo/LinksPage.tsx
@@ -17,13 +17,14 @@ const addFundsDefault = (() => {
 
 const LinksPage = () => {
   const demo = DemoContext.use();
+  const balance = demo.useBalance();
   const [, setPath] = usePath();
   const account = demo.useAccount();
 
   return (
     <div className="links-page">
       <Button
-        secondary
+        secondary={balance === undefined || balance > 0n}
         disabled={account === undefined}
         onPress={async () => {
           if (account === undefined) {
--- a/demos/inpage/demo/LinksPage.tsx
+++ b/demos/inpage/demo/LinksPage.tsx
@@ -78,6 +78,14 @@ const LinksPage = () => {
       )}
       <Button
         secondary
+        onPress={() => {
+          setPath('/registerAddress');
+        }}
+      >
+        Register Address
+      </Button>
+      <Button
+        secondary
         onPress={async () => {
           const granted = await demo.waxInPage.requestPermission(
             'This will reset the in-page wallet, including the stored private key. Are you sure?',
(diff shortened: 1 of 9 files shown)


JohnGuilding:
> Btw I forgot to mention that in a real product/deployment you'd probably proactively register the user's address without them doing it explicitly (and other common addresses). Another one of those things where we're showcasing dev tools instead of making a product. Curious to sync on that idea if anyone thinks differently.

I think implicitly doing things such as address registration could be nice. Even though we're showcasing dev tools, and not building a product, I think reducing the number of clicks and simplifying the demo could make it more impactful I.e. think there could be a benefit to showcasing what UX is possible in a product even though we're not building a product.

We could also tweak the flow so that the "register address" action is more meaningful. If you are a developer coming to the site and you see a button that says "register address", I'm not sure you would understand straight away what benefit that would have. Or when visiting the compression demo, what actions you should undertake to see all the compression savings. 

Regarding the compression demo generally, instead of displaying the bytes savings in the console, I think displaying the compression savings in the UI itself would highlight the benefits really well. Maybe this would involve simulating transactions with and without compression E.g. we could display a message along the lines of: "The calldata for this userOp was xxx bytes, an uncompressed and equivalent userOp would have taken up xxxx bytes. This is a reduction of xx%".

It could also be nice to represent that saving visually e.g. a bar chart showing calldata size with and without compression.

**TLDR**: think doing this sort of stuff implicitly is nice. If not, I think the flow could change slightly to showcase what registering an address means/what a user should do when they reach the compression demo to see the benefits. Whichever route we go down, I also think a UI component that shows the benefits of compression visually would be a good end result.

Let me know your thoughts :)

JohnGuilding (approved):
Some merge conflicts but looks good.

Left some thoughts on how we could showcase compression, but can be addressed in future work depending on what we decide for what works best

voltrevo:
Very minor merge conflict in package.json, going ahead with merge.

Issue #178, opened 2023-12-14
Detailed wax compression doc


voltrevo:
Share with 4337 mafia

voltrevo:
https://hackmd.io/@voltrevo/Bkz8syuUp

PR #179, opened 2023-12-15
Fee calculator

voltrevo:
## What is this PR doing?

Adds wax fee calculator subproject:
Screen Shot 2024-01-02 at 10 03 05

In `wax/demos/fee-calculator`, run `yarn dev`

Contributes to #178.

## Checklist

## Guidelines


demos/fee-calculator/src/App.tsx
```diff
@@ -0,0 +1,225 @@
...
+    defaults.bundlerProfitMargin,
+  );
+
+  const l1TransferFee = ethPrice * l1GasPrice * 1e-9 * constants.transferGas;
+
+  const l2TransferFee = (() => {
+    const l1Gas =
+      (l2CompressionRatio / calldataCostReduction4844) *
+      (constants.transferEffectiveBytes * constants.gasPerByte +
+        constants.l2FixedOverhead);
```
JohnGuilding, line 49: Can you explain this calculation?

demos/fee-calculator/src/App.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+    const l2Gas = constants.transferGas;
+
+    return ethPrice * 1e-9 * (l1GasPrice * l1Gas + l2GasPrice * l2Gas);
+  })();
+
+  const l2TransferFee4337 = (() => {
```
JohnGuilding, line 56: Trying to get my head around this expression as well. How did you arrive at the fixed values? (bundleOverheadEffectiveBytes - `155`, `455`, bundleOverheadGas - `25937`, `115195`)

demos/inpage/src/bundlers/SimulatedBundler.ts
```diff
@@ -35,6 +35,50 @@ export default class SimulatedBundler implements IBundler {
 
   constructor(waxInPage: WaxInPage) {
     this.#waxInPage = waxInPage;
+    (window as any).simulatedBundler = this;
+  }
+
+  async foo() {
```
JohnGuilding, line 41: Should this be named something else?

demos/inpage/src/WaxInPage.tsx
```diff
@@ -477,7 +478,9 @@ export default class WaxInPage {
         description,
         'is',
         hexLen(bytes),
-        'bytes:',
+        `bytes (${(Number(measureCalldataGas(bytes)) / 16).toFixed(
+          1,
+        )} effective):`,
```
JohnGuilding, line 483: What does effective mean here?

JohnGuilding (approved):
Nice looks good, left some questions on the calculations and a few others on inpage changes.

demos/inpage/src/bundlers/SimulatedBundler.ts
voltrevo: Indeed 😅. Will follow up.

demos/inpage/src/WaxInPage.tsx
voltrevo: I've been using it to mean equivalent non-zero bytes. So 4 zero-bytes equals one non-zero bytes since L1 charges 16 gas for non-zeros and 4 gas for zeros.

demos/fee-calculator/src/App.tsx
voltrevo: Sure. The starting point is `constants.transferEffectiveBytes * constants.gasPerByte`. This is a crude calculation of the amount of L1 gas that the L2 needs to charge for.

Multiplying this by `l2CompressionRatio / calldataCostReduction4844` accounts for L2-level compression (gzip/brotli/etc) and the future scale change that blobs (4844) will introduce. Currently we just default `calldataCostReduction4844` to 1 so that it has no impact, because we don't have blobs yet.

Finally, there's an additional overhead that needs to get scaled in the same way. I imagine this exists because of extra metadata that needs to be posted to L1, and the L2 is amortizing this cost over each transaction equally.

This maps exactly onto Optimism's model if you use the right parameters.

I'm planning to change the model slightly to reflect this work I've been doing to [document wax compression](https://hackmd.io/lFYwpXVWS6iJ6w7HnYM0YQ?view), because it lends itself to direct measurement instead of knowing the internals of the L2:

Screen Shot 2024-01-09 at 10 21 24

...

Screen Shot 2024-01-09 at 10 23 01

demos/fee-calculator/src/App.tsx
voltrevo: Yeah sorry about that. I need to clean this up a bit.

Anyway, 455 is the number of (effective) bytes in a bundle with one transfer. Therefore, including a transfer adds `455 - 155 = 300` bytes to the bundle. The bytes needed for a bundle even when it's empty need to be shared across all the user ops, which is why `effectiveBytesCharged` is `bundleOverheadEffectiveBytes / bundleSize + addedEffectiveBytes`.

`bundleOverheadGas` and `addedGas` are the same idea but for the gas native to the L2 (l2Gas). In the same way, an empty bundle needs a certain amount of gas, and then userOps expand that gas, so we charge fully for the added gas, and share the overhead needed for the bundle.

The constants involved for these are based on measurements taken using the in-page demo.

See previous answer for `l1Gas` calculation.

demos/fee-calculator/src/App.tsx
JohnGuilding: Thanks for the explanation - makes sense now.

Only follow up question I have is if `l2CompressionRatio` accounts for the compression done by the L2, how did you arrive at the value of `0.7`? Was it an estimate given by optimism/arbitrum

demos/fee-calculator/src/App.tsx
JohnGuilding: Thanks for the clarification, all makes sense

demos/fee-calculator/src/App.tsx
voltrevo: It's based on optimism since they give an explicit value on-chain and in the block explorer of 0.684. I expected arbitrum to give similar results since brotli/gzip are fairly similar. Further testing seems to show optimism giving better results here and I'm not sure why.

Issue #181, opened 2024-01-05
Email Recovery Timelock/Delay


jacque006:
*As a* smart contract account owner using email recovery
*I would like to* Have a delay between when I being the recovery process and when it completes (timelock/delay)
*So I can* Ensure that I can cancel the recovery process if my email is temporarily compromised 

### Details

- Make delay configurable
- Set the default to a long time period (2 weeks?) or don't have one to ensure consumers think carefully about what they want to set it to

PR #183, opened 2024-01-09
Add default and custom delay to zkemail recovery

JohnGuilding:
## What is this PR doing?
- Adds a default delay of 2 weeks for recovery
- Adds a configurable custom delay for recovery which overrides the default delay
- Adds a cancel recovery function. Access control for this function relies on the fact that msg.sender is the safe for which recovery should be cancelled for. We may want to add more robust access control in future work.

### Basic flow
3 functions now need to be called in the recovery plugin to successfully rotate an owner on the ecsda plugin: `configureRecovery`, `initiateRecovery` & `recoverPlugin`. Also `setCustomDelay` if setting a custom delay.

- `configureRecovery` (previously called `addRecoveryHash`) largely does what it did before, which is storing a recovery request in the form of a `recoveryHash` and a `dkimPublicKeyHash` (other values not set on the struct until later). Designed to be called by the safe via use of msg.sender
- `initiateRecovery` starts the recovery process and verifies the zkp required to recover the account. `executeAfter` and `pendingNewOwner` are then set as part of the recovery request. The account can then be recovered after the delay has passed. Currently designed to be called by anyone as long as the proof and other arguments are valid.
- `recoverPlugin` can be called by anyone and executes the recovery if the delay has elapsed. This function and also `initiateRecovery` replaced the previous function that executed recovery called `recoverAccount`.

### Why do we need a default delay _AND_ a custom delay?
A default delay provides an initial sensible default that follows best practises. This value should aim to provide users with a delay that provides them enough time to react to malicious/accidental recoveries, while not being too long as to leave them waiting for a long time before they can use their account again. However, different individuals/organisations may have differing requirements that mean a value such as 48 hours is far too long for the low value stored in the account. Alternatively, 48 hours may be too short to account for time spent on holiday/away from work. A custom delay allows the end user to specify a delay that works best for their own security and context.

Note: The default delay has been set to a high value of 2 weeks on purpose which might not be classed as a _sensible_ default. This is because we want developers who use our code to understand it and configure it to suit their specific context, as opposed to blindly copying and pasting (based off of learnings from other PSE projects). 

Run integration tests

#181

## Checklist

## Guidelines

### DIFF
--- a/account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
+++ b/account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
@@ -217,7 +217,6 @@ contract SafeZkEmailRecoveryPlugin {
         bool verified = verifier.verifyProof(a, b, c, publicSignals);
         if (!verified) revert INVALID_PROOF();
 
-        // FIXME: if recovery is not configured, should not get here
         uint256 executeAfter = block.timestamp + recoveryDelay[safe];
 
         recoveryRequests[safe].executeAfter = executeAfter;
--- a/account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
+++ b/account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
@@ -10,9 +10,11 @@ interface ISafeECDSAPlugin {
     function getOwner(address safe) external view returns (address);
 }
 
-struct ZkEmailRecoveryStorage {
+struct RecoveryRequest {
     bytes32 recoveryHash;
     bytes32 dkimPublicKeyHash;
+    uint256 executeAfter;
+    address pendingNewOwner;
 }
 
 /**
@@ -20,25 +22,34 @@ struct ZkEmailRecoveryStorage {
  * NOT FOR PRODUCTION USE
  */
 contract SafeZkEmailRecoveryPlugin {
-    // Default DKIM public key hashes registry
+    /** Default DKIM public key hashes registry */
     IDKIMRegsitry public immutable defaultDkimRegistry;
+
+    /** Default delay has been set to a large timeframe on purpose. Please use a default delay suited to your specific context */
+    uint256 public constant defaultDelay = 2 weeks;
 
     bytes32 immutable RECOVERY_HASH_DOMAIN;
 
-    // Mapping of safe address to plugin storage
-    mapping(address => ZkEmailRecoveryStorage) public zkEmailRecoveryStorage;
+    /** Mapping of safe address to recovery request */
+    mapping(address => RecoveryRequest) public recoveryRequests;
 
-    // Mapping of safe address to dkim registry address
+    /** Mapping of safe address to a custom recovery delay */
+    mapping(address => uint256) public customDelay;
+
+    /** Mapping of safe address to dkim registry address */
     mapping(address => address) public dkimRegistryOfSafe;
 
+    error MODULE_NOT_ENABLED();
+    error INVALID_OWNER(address expectedOwner, address owner);
+    error RECOVERY_ALREADY_INITIATED();
     error INVALID_DKIM_KEY_HASH(
         address safe,
         string emailDomain,
         bytes32 dkimPublicKeyHash
     );
-    error MODULE_NOT_ENABLED();
-    error INVALID_OWNER(address expectedOwner, address owner);
     error INVALID_PROOF();
+    error RECOVERY_NOT_INITIATED();
+    error DELAY_NOT_PASSED();
 
     MockGroth16Verifier public immutable verifier;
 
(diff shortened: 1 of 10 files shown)


account-integrations/safe/test/forge/SafeZkEmailRecoveryPlugin.t.sol
```diff
@@ -148,23 +221,32 @@ contract SafeZkEmailRecoveryPluginTest is TestHelper {
 
         // Act
         vm.startPrank(safeAddress);
-        safeZkEmailRecoveryPlugin.addRecoveryHash(
+        vm.expectEmit(true, true, false, false);
```
JohnGuilding, line 224: See the [foundry cheatcode docs](https://book.getfoundry.sh/forge/cheatcodes?highlight=expectEmit#cheatcodes) for more info on how `expectEmit` works

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
```diff
@@ -94,69 +126,146 @@ contract SafeZkEmailRecoveryPlugin {
...
+     */
+    function recoverPlugin(address safe, address ecdsaPlugin) external {
+        RecoveryRequest memory recoveryRequest = recoveryRequests[safe];
+
+        if (recoveryRequest.executeAfter == 0) {
+            revert RECOVERY_NOT_INITIATED();
+        }
+
+        if (block.timestamp > recoveryRequest.executeAfter) {
+            delete recoveryRequests[safe];
```
JohnGuilding, line 229: Was torn on whether we should delete the mapping value after recovering the plugin. Curious if anyone had opinions on whether this was worthwhile.

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
```diff
@@ -79,7 +111,7 @@ contract SafeZkEmailRecoveryPlugin {
      * @param dkimPublicKeyHash hash of DKIM public key - keccak256(abi.encodePacked(dkimPublicKey))
      * @param dkimRegistry address of a user-defined DKIM registry
      */
-    function addRecoveryHash(
+    function configureRecovery(
```
jacque006, line 114: Good to have an optional field here to set a custom delay as part of the initial recovery config. Saves a txn/UserOp if someone wants a custom delay.

I would still keep `setCustomDelay` in case someone wants to update it, such as in your example if they go on vacation or are away for some reason.

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
```diff
@@ -94,69 +126,146 @@ contract SafeZkEmailRecoveryPlugin {
...
-        bytes memory data = abi.encodeWithSignature(
-            "enable(bytes)",
-            abi.encodePacked(newOwner)
-        );
+        uint256 delay;
+        if (customDelay[safe] > 0) {
+            delay = customDelay[safe];
+        } else {
+            delay = defaultDelay;
+        }
```
jacque006, line 206: Move this into `configureRecovery` and always set the delay for each config to be explicit. You might also want to rename the mapping to something like `recoveryDelay`.

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
jacque006: The question is, would someone want to use the same recovery configuration across multiple recoveries? I think the answer is yes. You should be able to recover using a specific email as many times as you want until such time as that recovery configuration is updated or removed. This is very similar to how existing password resets work: You can reset as many times as you want with an email until you update that email.

This also brings up another interesting question which is: Should we have a delay on adding, updating, or removing the recovery config as well. There could be cases where depending on the access restrictions/pattern we have on modifying the recovery config an attacker could update the config as the first txn of the attack, rendering the recovery useless. This also might make it worthwhile to think about locking/freezing the safe while the recovery process is ongoing.

I think all of the above is out of scope for this work, but is worth considering for future improvements to harden this recovery process. I would leave this as-is.

jacque006 (commented):
Some minor changes that would be good to see, otherwise looks good.

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
JohnGuilding: Yeah good point. Added some notes on recovery hardening to a new issue that can be broken down https://github.com/getwax/wax/issues/184

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
JohnGuilding: Added

account-integrations/safe/src/SafeZkEmailRecoveryPlugin.sol
JohnGuilding: makes sense, done

jacque006 (approved):

Issue #173, opened 2023-12-09
Create Email Recovery Relayer


jacque006:
*As a* Smart contract account owner using email recovery
*I would like to* Be able to generate the email proof needed for recovering my account
*So I can* Recover my account using local tools, or rely on a 3rd party running the safe software

### Details
- Core can just be a simple Typescript (TS) HTTP server with two endpoints:
  - `/prove` or similar: Generates the email recovery proof
  - `/health` or similar: Check liveness of server
- Needs to be able to receive emails. Unsure how to structure this part of workflow
  - Check with ZKEmail team how these did this for their current relayer 

### Considerations
- Should this just be a TS script/lib that others can integrate into a server/local script?
- Do we need to manage/have a queue of proofs & the ability to check their status?
  - Individual proof generation would take quite some time

jacque006:
Work completed. See linked PRs

PR #200, opened 2024-01-29
Process emails and initiate recovery

JohnGuilding:
This PR is dependent on https://github.com/getwax/wax/pull/199, merge that first

## What is this PR doing?

- Create sqlite db and save emails
- Mocks processing of emails
- Initiates recovery on recovery plugin 
- Adds script to set up recovery for relayer - `setupRecoveryForRelayer.ts`
- Adds basic formatting and linting

## Happy path flow
1. Once an email is fetched via `EmailService.pollEmails()`, the app saves it to a the db and emits and event.
2. This event will trigger `EmailService.processEmails()` which finds eligible emails from the db, generates the circuit inputs (mocked for now), and calls the ethereum service.
3. `EthereumService.initiateRecovery()` is called which executes the `initiateRecovery` function on the recovery plugin
4. The email service will then mark the email as processed in the db

Testing these changes is quite manual at the moment, can definitely streamline this in future work. Instead of running this manually yourself, you can watch a demo [here](https://www.loom.com/share/b0829cc3b933420495756b25937ea307?sid=0ad92f1d-4467-49da-966d-f2f5a0d52650).

1. Run `yarn setup` in `/zkp/relayer` and add your email details to the config file - I created a new gmail account specifically for testing. Creating an app password was the quickest option to get things working. You can google how to do that but you need to set up mfa before you can generate one. You the generated app password as the password for your gmail account in the config file.
2. Start a local hardhat node with `./script/start-hardhat.sh` in `account-integrations/safe`
3. Start a local bundler. `yarn run bundler --unsafe` if using eth infinitisms (need to do this automatically in previous script)
4. Set up recovery by running `yarn hardhat run ./script/setupRecoveryForRelayer.ts` in `account-integrations/safe`.
5. Send an email to your chosen email address (subject etc doesn't matter for now - parsing the email fully will be an upcoming PR)
6. Copy `SafeZkEmailRecoveryPlugin.json` from `account-integrations/safe/out` (might need to build contracts with `forge build`), paste it in `account-integrations/zkp/relayer/src/services/`. This is for the abi.
7. Start the node app with `yarn dev` in `zkp/relayer`. The email should be picked up and recovery should be initiated automatically. You should see this log printed `Recovery initiated by: ${emails[i].sender}`

Contributes to #173

## Checklist

## Guidelines


account-integrations/zkp/relayer/src/services/emailService.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    async stop() {
+        await this.imapClient.stop();
+    }
+
+    public async pollEmails(): Promise<void> {
+        await this.imapClient.start();
+        // eslint-disable-next-line no-constant-condition
+        while (true) {
```
jacque006, line 29: I would instead include some termination flag for this polling loop, such as:
```suggestion
        while (running) {
```
Can set to `true` when `start` is called, and set to `false` when `stop` is called.

jacque006 (approved):

account-integrations/zkp/relayer/src/services/emailService.ts
JohnGuilding: Follow up https://github.com/getwax/wax/pull/206

PR #202, opened 2024-01-31
BLS plugin: use aggregation

voltrevo:
## What is this PR doing?

Updates `SafeBlsPlugin` to use aggregation via eth-infinitism's `BLSSignatureAggregator` example.

Supporting changes:
- Switch to [arachnid's deterministic deployer](https://github.com/Arachnid/deterministic-deployment-proxy) for non-safe contracts for consistency with eth-infinitism and the broader ecosystem[^1][^2]
- Add support for linked libraries to deterministic deployer
- Patch the bundler to add partial aggregation support

Run the tests per instructions ([allowing the webauthn test to fail](https://github.com/getwax/wax/issues/191)).

Contributes to #190.

## Checklist

## Guidelines

[^1]: [Arachnid's deployer](https://etherscan.io/address/0x4e59b44847b379578588920ca78fbf26c0b4956c) is >10x more popular than [safe's deployer](https://etherscan.io/address/0x914d7Fec6aaC8cd542e72Bca78B30650d45643d7)
[^2]: Safe originally forked Arachnid's deployer so that they could use a couple of chains that were incompatible at the time. However, those chains have since added the necessary support for legacy transactions, so this foundational reason for Safe's deployer, which requires ongoing cooperation from a specific third party, no longer applies.

### DIFF
--- a/account-integrations/safe/script/patched-bundler-727838b.dockerfile
+++ b/account-integrations/safe/script/patched-bundler-727838b.dockerfile
@@ -1,0 +1,16 @@
+FROM accountabstraction/bundler:0.6.2
+
+RUN apt-get update && \
+    apt-get install -y coreutils && \
+    rm -rf /var/lib/apt/lists/*
+
+ARG EXPECTED_DIGEST="727838bd8705dd319970fb4a86c9abe5334e17312230143e01043abf0732b5f7"
+
+# Modify /app/bundler.js to add support for estimating (but not running) aggregate bundles
+RUN sed -i 's/(errorResult.errorName !== '\''ValidationResult'\'')/(!errorResult.errorName.startsWith('\''ValidationResult'\''))/g' /app/bundler.js
+
+# Fail if we didn't achieve the expected SHA256 digest
+RUN if [ "$(sha256sum /app/bundler.js | awk '{print $1}')" != "$EXPECTED_DIGEST" ]; then \
+    echo "SHA256 digest does not match. Aborting."; \
+    exit 1; \
+fi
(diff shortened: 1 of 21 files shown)


account-integrations/safe/script/patched-bundler-727838b.dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ARG EXPECTED_DIGEST="727838bd8705dd319970fb4a86c9abe5334e17312230143e01043abf0732b5f7"
+
+# Modify /app/bundler.js to add support for estimating (but not running) aggregate bundles
+RUN sed -i 's/(errorResult.errorName !== '\''ValidationResult'\'')/(!errorResult.errorName.startsWith('\''ValidationResult'\''))/g' /app/bundler.js
+
+# Fail if we didn't achieve the expected SHA256 digest
+RUN if [ "$(sha256sum /app/bundler.js | awk '{print $1}')" != "$EXPECTED_DIGEST" ]; then \
+    echo "SHA256 digest does not match. Aborting."; \
+    exit 1; \
+fi
```
jacque006, line 16: I think a better way to do this would be to maintain a fork of `accountabstraction/bundler` with these changes, This would make the needed changes more visible, make it easier to update with future versions, and could eventually also be a place we add needed changes to get compression working. This would also make it easier for us to contribute these changes back the the main repo. Out of scope for this work.

jacque006 (approved):
:+1: Also wanted to give a shout out to you & @JohnGuilding for the recent verbose PR descriptions & code comments. Makes reviewing these much easier, and sets a good example for me to follow with my own contributions :smile:

PR #204, opened 2024-02-05
Pull new emails

JohnGuilding:
This PR is dependent on https://github.com/getwax/wax/pull/201, merge that first

## What is this PR doing?
- Pulls new emails while the app is running - before the app could not detect new emails in an inbox. Only emails already in the inbox would be picked up
- Copies contract artifacts from contracts directory on `yarn setup`

n/a

contributes to #173 

## Checklist

## Guidelines

### DIFF
--- a/account-integrations/zkp/relayer/src/lib/imapClient.ts
+++ b/account-integrations/zkp/relayer/src/lib/imapClient.ts
@@ -1,41 +1,62 @@
 import { ImapFlow, ImapFlowOptions } from "imapflow";
 
 class ImapClient {
-    public imapConfig: ImapFlowOptions;
+    public imapClient: ImapFlow;
 
-    constructor(imapConfig: ImapFlowOptions) {
-        this.imapConfig = imapConfig;
+    constructor(imapClientConfig: ImapFlowOptions) {
+        this.imapClient = new ImapFlow(imapClientConfig);
+    }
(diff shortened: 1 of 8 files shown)


account-integrations/zkp/relayer/src/lib/imapClient.ts
```diff
@@ -4,55 +4,61 @@ type EmailResponse = {
...
-                    headers: true,
-                    envelope: true,
-                    source: true,
-                    bodyStructure: true,
-                    flags: true,
-                }
-            );
+            // For some reason calling .status() seems to "refresh" the inbox so that
+            // new emails can be detected. Without this line, new emails are not detected.
+            const mailbox = await this.imapClient.status("INBOX", {
```
JohnGuilding, line 32: This is the line that I added that enabled the imap client to fetch new emails. `this.imapClient.fetch` on its own did not suffice. 

Another approach I tried that worked was to persist an imapClient instance over the course of this function. E.g. `const imapClient = new ImapFlow(imapClientConfig)`, and then start and end the connection in this function. It felt better to connect to an inbox once and persist the connection, rather than connecting and logging out everytime the inbox is being polled - you can see this approach in the `Receive new emails` commit.

jacque006 (approved):

PR #209, opened 2024-02-21
192 restructure monorepo

### LINKED ISSUE #192: Restructure monorepo
The `wax` monorepo has lot of directories that might not be set out in the most optimal way. This makes the repo more complicated than it needs to be and is less contributor-friendly. 

 ### Potential changes
- Unify contracts
  - Could use duel hardhat/foundry project to hold additional contracts. For example, some contracts are in `/primitives` and `/kernel` which we may want to move over
- Decide if ethdk directory is needed - currently sitting empty with no upcoming ethdk work

JohnGuilding:
## What is this PR doing?
Restructures the monorepo.

- Rename `account-integrations` to `packages`
- Move `demos` into `packages`
- Rename `safe` to `plugins`
- Move `kernel`, `primitives` into `plugins` to utilise shared foundry & hardhat project
- Update test directory structure in `plugins` to account for new changes
- Had some submodule difficulties when migrating `kernel` so updated the contracts to use latest `kernel` release as that mitigated the submodule issue I was getting. As part of this had to generate new bls sigs, so added a script to do that - `plugins/script/generateBlsSig.ts`
- Delete ethdk - there is currently no code in this directory and we have no upcoming work related to this. So thought removing would make the repo simpler to understand.
- Updates READMEs
- Found an issue with the bls e2e tests not related to this PR - raised the following issue https://github.com/getwax/wax/issues/210

Feel free to push back on any of the above changes. Also if I missed any worthwhile changes, feel free to suggest them.

In terms of testing locally, all the tests pass as they did before (except the one mentioned, but that looks unrelated as was failing in a commit before these changes), workflows are passing in CI, and I also tried out the demos and they seem to work as before. I didn't try everything in the inpage, but tested we could still deploy an account, send a tx and call a contract.

## Future/other potential changes
- While walking through the relayer code with Yush from zkemail, he thought keeping the recovery contracts like `SafeZkEmailRecoveryPlugin.sol` next to the relayer code would make it easier for developers to understand how recovery works and what contracts are relevant. Haven't made this change but mentioning the idea here for reference.
- While the READMEs have been updated to account for these changes, they probably still require updating fully to account for current WAX priorities and changes to the monorepo. README changes in this PR make sure it's not completely outdated. Added an issue to update these properly #211 
- This PR adds no monorepo tooling, and is primarily concerned with restructuring the directories. If we want to add some more sophisticated tooling, the idea is that could be covered by a future PR.
- Not sure what the plans is with the sdk and barebones contracts - left them as they are for now

- Examine directory structure locally
- Run e2e tests

Resolves #192 

## Checklist

## Guidelines


jacque006 (changes requested):
I'm seeing the compression project twice in this project setup:
- Once in `./packages/compression`
- Again in `./packages/plugins/src/compression`

I assume it should only be in the top level `packages` folder?

Otherwise, everything else looks good :+1:

jacque006 (approved):

PR #232, opened 2024-04-09
Update commits with base branch & add new app context with calls

jacque006:
More integration work for email recovery demo


jacque006:
FYI @Rohan-cp . I still need to test new context & funcs

jacque006:
Discarding in favor of other UI

